PR #2, opened 2012-07-22
Enhancements

ajshort:
* Add an index search filter.
* Fix request processor definition syntax.


chillu:
I've checked each of the commits, and the underlying code has changed substantially - looks like they no longer apply. Very sorry for the late response Andrew :(

PR #42, opened 2014-05-07
added better stemming support

pitchandtone:
Based on http://wiki.apache.org/solr/AnalyzersTokenizersTokenFilters#Stemming

### DIFF
--- a/code/solr/Solr.php
+++ b/code/solr/Solr.php
@@ -173,7 +173,7 @@ class Solr_BuildTask extends BuildTask
 
     /**
      * Get the current logger
-     * 
+     *
      * @return LoggerInterface
      */
     public function getLogger()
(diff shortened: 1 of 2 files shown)


chillu:
Thanks John! Looks like we should just cherry pick 791c2f965d7822b087062a4873d1dd25bf8090df from this pull request, right? The other commits look unrelated and unmergeable.

@hafriedlander This only works with Solr 4.3 onwards, which was released over two years ago. I know we have a Solr3Service class in the codebase, but don't declare specific Solr version support anywhere. Should we drop support for Solr releases prior to 4.3? What version are we running in CWP?

pitchandtone:
That's correct. Note that the update only applies to the solr 4 conf file so won't break anyone running 3.

pitchandtone:
Do you want me to redo this PR will just that commit? Or happy to cherry pick @chillu @hafriedlander ?

dhensby:
@pitchandtone If you can open a new PR with just that commit, that'd be most helpful to allow discussion around it, thanks

chillu:
Yeah agree that a new PR is easier - it means we can see the tests passing without having to run them locally.

PR #46, opened 2014-06-03
Trigger update for publish and unpublish

cameronbourgeois:


dhensby:
@tractorcow how do we feel about this PR? Do we want to get it in. Is it still a  problem that unpublished content remains in the index?

tractorcow:
Missing check for `$this->owner->ID` being empty. It would be good if there were unit tests to show what this actually fixed.

tractorcow:
Closed in favour of https://github.com/silverstripe/silverstripe-fulltextsearch/pull/137#issuecomment-284604644, which fixes the issue a bit more directly. :)

tractorcow:
I repent. This PR actually introduces the necessary onAfter* hooks for catching item deletes.

tractorcow:
Apparently, this code is preventing delete manipulations being processed.

```
// Trim records without fields
        foreach(array_keys($writes) as $key) {
            if(empty($writes[$key]['fields'])) {
                unset($writes[$key]);
            }
        }
```

Because deletes don't have a fields key.

tractorcow:
Replaced with https://github.com/silverstripe/silverstripe-fulltextsearch/pull/138

PR #52, opened 2014-07-16
Added support for Translatable

Leapfrognz:
Adds solr support for translatable (uses multiple cores).
Set the locale for each solr index (which extend SolrIndex) in config yml:

ChineseSiteSearchIndex:
  limitToLocale: zh_CN
SiteSearchIndex:
  limitToLocale: en_NZ

It will then append the correct locale to the cmd.

I also added the ability to quickly run a simple reindex on a particular index:
sake dev/tasks/Solr_Reindex index=ChineseSiteSearchIndex verbose=1

### DIFF
--- a/code/solr/Solr.php
+++ b/code/solr/Solr.php
@@ -206,13 +206,20 @@ class Solr_Reindex extends BuildTask {
 		increase_time_limit_to();
 		$self = get_class($this);
 		$verbose = isset($_GET['verbose']);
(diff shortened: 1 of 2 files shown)


chillu:
Sorry for the late reply, but I think this isn't the best approach. It should be implemented as a `SearchVariant`, similar to `SearchVariantSubsites`.

PR #61, opened 2015-02-16
BUG Add missing addStoredFields method

tractorcow:
This has been documented in the readme for ages but never implemented.

Tested with localsolr with highlighting enabled.

cc @hafriedlander


tractorcow:
cc @chillu

chillu:
Thanks Damian, that was my bad around "docs without implementation" (78c5daa18613c04a6c0401b2257c101ab8f4631d). It's been a while, but from memory I've untangled a stream of work into half a dozen pull requests which partially relied on each other, which is why this one slipped through.

PR #62, opened 2015-05-04
Docs on index triggers, queues and timeouts

chillu:
If somebody has more info on PHP timeout/abort behaviour in various web server implementations (mod_fcgi, mod_php, nginx, IIS), please add here.

cc @hafriedlander @tractorcow


dhensby:
I merged what I could of this by hand.

PR #64, opened 2015-05-05
Clean up SearchQuery API (less misleading names, fluent interface)

chillu:
`SearchQuery->search()` is highly misleading - it implies that a search is executed,
while in fact it only adds a search term to an internal property.
This gets even more confusing since the actual method executing passing
on the search request to the backend is also called `SearchIndex->search()`.
And many implementations also create a `Page_Controller->search()` one layer further up.

Renamed other methods along the same lines, for example `start()` is a confusing
name for adding a start value for pagination, and `inClass()` implies a boolean
return while in fact its adding a filter.

Also added getters to get naming a bit more consistent (e.g. `filter()` sets `$require`).

Set the deprecation notices to 2.0, so a future release.

### DIFF
--- a/code/search/SearchQuery.php
+++ b/code/search/SearchQuery.php
@@ -8,6 +8,7 @@
 class SearchQuery extends ViewableData {
 
 	static $missing = null;
+	
 	static $present = null;
 
 	static $default_page_size = 10;
@@ -19,12 +20,12 @@ class SearchQuery extends ViewableData {
 	public $classes = array();
 
 	public $require = array();
+
 	public $exclude = array();
 
 	protected $start = 0;
-	protected $limit = -1;
 
-	/** These are the API functions */
+	protected $limit = -1;
 
 	function __construct() {
 		if (self::$missing === null) self::$missing = new stdClass();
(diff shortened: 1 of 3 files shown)


code/search/SearchQuery.php
```diff
@@ -77,32 +100,116 @@ function filter($field, $values) {
...
 		$this->limit = $limit;
+		return $this;
 	}
 
-	function page($page) {
+	function getLimit() {
+		return $this->limit;
+	}
+
+	function setPageSize($page) {
```
tractorcow: I think these methods need `public ` and some phpdoc. I'm not sure what this is doing immediately by reading the code.

code/search/SearchQuery.php
```diff
@@ -77,32 +100,116 @@ function filter($field, $values) {
...
 		$this->start = $page * self::$default_page_size;
 		$this->limit = self::$default_page_size;
+		return $this;
+	}
+
+	public function getPageSize() {
```
tractorcow: Why is this an alias for getLimit()? What's the distinction?

dhensby:

andrewandante:
@chillu as part of the ORB work, I'm considering integrating equivalent changes as a way of reducing developer confusion when working with the module. If you're happy for me to lift and tweak I will do so 😄

chillu:
> If you're happy for me to lift and tweak I will do so

Yes please! Thank you Andrew :)

dhensby:
@chillu can you give contributors push access to your PR branch, please :)

dhensby:
I've rebased this, just need to be able to push it

dhensby:
oh, looks like @andrewandante has already done it in #214

PR #66, opened 2015-05-07
Remove child process model in Solr_Reindex

chillu:
Ran a benchmark on PHP 5.3 with 6k pages and 20k files (around ~300MB of index size overall). Memory usage increased from 53MB to 62MB, which IMHO is an acceptable leakage. That'll leave us a lot of room until hitting common limits like 128MB

### DIFF
--- a/code/search/SearchIndex.php
+++ b/code/search/SearchIndex.php
@@ -367,7 +367,7 @@ abstract class SearchIndex extends ViewableData {
 	 * @return Mixed - The value of the field, or null if we couldn't look it up for some reason
 	 */
 	protected function _getFieldValue($object, $field) {
-		set_error_handler(create_function('$no, $str', 'throw new Exception("HTML Parse Error: ".$str);'), E_ALL);
+		set_error_handler(array($this, '_handleError'), E_ALL);
 
 		try {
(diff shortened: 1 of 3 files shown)


chillu:
Note: Hamish wants tests on 100k pages being indexed. Can provide that, but not right now :)

tractorcow:
Hi @chillu , we have superceded this work with a new re-index task as well as a new queuedjobs processor. You are welcome to rebase so that the default processer no longer uses the php background task, although the actual invocation is now in a new class.

You are able to in fact override this with an alternative subclass, since the actual runner is now DI, giving you the option to choose between the old (background process) or your suggested one (runs directly in the same process).

See https://github.com/silverstripe-labs/silverstripe-fulltextsearch/blob/master/code/solr/reindex/handlers/SolrReindexImmediateHandler.php and https://github.com/silverstripe-labs/silverstripe-fulltextsearch/blob/master/code/solr/reindex/handlers/SolrReindexBase.php

Your `processGroup` method would simply call `runGroup` in your case. :D

dhensby:
@chillu - do you have any desire to revisit this?

robbieaverill:

PR #67, opened 2015-05-07
API Separate searchupdate / commit into separate queued-jobs

tractorcow:
Oh my good, look at this pull request. It's so amazing!

What have I done! On a coffee-less afternoon nonetheless?

- Indexes can finally hide other indexes. This is useful when you don't want classes to be indexed, but they could be in a third party module you don't want to edit. (see docs below) This in itself is a great performance increase, but no longer do you need to have solr indexing work done on indexes you don't want to maintain.

- SolrIndex::commit has been split into a separate queued job which is added to the queue once an update job has been completed. This is useful if you want to run multiple update jobs in a sequence, but don't want to run into the dreaded `exceeded limit of maxWarmingSearchers=X` error.

It also nicely bunches several smaller updates into a single commit, so hopefully less work is done overall.

Also if for whatever reason a commit is made in a single job, and updates occur later (again, requiring further updates) then the job will be re-added to the queue with a nice cooldown (defaults to 10 minutes) to prevent solr erroring, and giving time for the rest of the job queue to complete.

### DIFF
--- a/code/search/FullTextSearch.php
+++ b/code/search/FullTextSearch.php
@@ -10,6 +10,15 @@ class FullTextSearch {
 	static protected $indexes_by_subclass = array();
 
 	/**
+	 * Optional list of index names to limit to. If left empty, all subclasses of SearchIndex
+	 * will be used
+	 *
+	 * @var array
+	 * @config
+	 */
+	private static $indexes = array();
+
+	/**
 	 * Get all the instantiable search indexes (so all the user created indexes, but not the connector or library level
 	 * abstract indexes). Can optionally be filtered to only return indexes that are subclasses of some class
 	 *
@@ -18,16 +27,43 @@ class FullTextSearch {
 	 * @param bool $rebuild - If true, don't use cached values
 	 */
 	static function get_indexes($class = null, $rebuild = false) {
-		if ($rebuild) { self::$all_indexes = null; self::$indexes_by_subclass = array(); }
+		if ($rebuild) {
+			self::$all_indexes = null;
+			self::$indexes_by_subclass = array();
+		}
 
 		if (!$class) {
 			if (self::$all_indexes === null) {
-				$classes = ClassInfo::subclassesFor('SearchIndex');
+				// Get declared indexes, or otherwise default to all subclasses of SearchIndex
+				$classes = Config::inst()->get(__CLASS__, 'indexes')
+					?: ClassInfo::subclassesFor('SearchIndex');
 
-				$concrete = array();
+				$hidden = array();
+				$candidates = array();
 				foreach ($classes as $class) {
+					// Check if this index is disabled
+					$hides = $class::config()->hide_ancestor;
+					if($hides) {
+						$hidden[] = $hides;
+					}
+
+					// Check if this index is abstract
 					$ref = new ReflectionClass($class);
-					if ($ref->isInstantiable()) $concrete[$class] = singleton($class);
+					if (!$ref->isInstantiable()) {
+						continue;
+					}
+
+					$candidates[] = $class;
+				}
+
+				if($hidden) {
+					$candidates = array_diff($candidates, $hidden);
+				}
+
+				// Create all indexes
+				$concrete = array();
+				foreach($candidates as $class) {
+					$concrete[$class] = singleton($class);
 				}
 
 				self::$all_indexes = $concrete;
@@ -60,6 +96,8 @@ class FullTextSearch {
 	 * From then on, fulltext search system will only see those indexes passed in this most recent call.
 	 *
 	 * Passing in no arguments resets back to automatic index list
+	 *
+	 * Alternatively you can use `FullTextSearch.indexes` to configure a list of indexes via config.
 	 */
 	static function force_index_list() {
 		$indexes = func_get_args();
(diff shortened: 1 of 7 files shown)


code/search/FullTextSearch.php
```diff
@@ -18,16 +18,41 @@ class FullTextSearch {
...
 		if (!$class) {
 			if (self::$all_indexes === null) {
 				$classes = ClassInfo::subclassesFor('SearchIndex');
 
-				$concrete = array();
+				$hidden = array();
+				$candidates = array();
 				foreach ($classes as $class) {
+					// Check if this index is disabled
+					$hides = $class::config()->hide_ancestor;
```
chillu: I would prefer to make this an explicit registration system, less magic and more versatile. The YAML configuration doesn't work because we can't overwrite lists, but we could use `Config::update()` for that?

```yml
FullTextSearch:
  indexes:
    - MySolrIndex
```

phptek:
@tractorcow If this works I will have your babies.

code/search/FullTextSearch.php
tractorcow: I wanted to go for something that, if ignored, doesn't have any effect, and is fully backwards compatible.

You can do the same in yaml though.

```
MyNewIndex:
  hide_ancestor: MyOldIndex
```

Ok, maybe I can split it into a new PR.

tests/BatchedProcessorTest.php
```diff
@@ -100,7 +138,60 @@ public function testBatching() {
...
+		$first = SearchUpdateCommitJobProcessor::queue();
+		$second = SearchUpdateCommitJobProcessor::queue();
+
+		$this->assertFalse($index->getIsCommitted());
+
+		// First process will cause the commit
+		$this->assertFalse($first->jobFinished());
+		$first->process();
+		$allMessages = $first->getMessages();
+		$this->assertTrue($index->getIsCommitted());
```
patbolo: Shouldn't the commit be performed after $second has been processed?

tractorcow: If the first commit has run, it should be committed then, not after the `$second->process()`.

When the second one is run it should see that there is in fact no more committing left to do, hence the "discarding this job" message.

tests/BatchedProcessorTest.php
```diff
@@ -100,7 +138,60 @@ public function testBatching() {
...
+		$this->assertTrue($first->jobFinished());
+		$this->assertStringEndsWith('All indexes committed', $allMessages[2]);
+
+		// Executing the subsequent processor should not re-trigger a commit
+		$index->reset();
+		$this->assertFalse($second->jobFinished());
+		$second->process();
+		$allMessages = $second->getMessages();
+		$this->assertFalse($index->getIsCommitted());
```
tractorcow: @patbolo I only temporarily reset the $index to set comitted to false to prove that it's not re-comitted by `$second`. In practice it would stay comitted, although just for the sake of this test we toggle it back to false.

normann:
@phptek lust comment

code/search/FullTextSearch.php
tractorcow: Ok I'm going to add both; You can use an 'indexes' config to do a pre-filter, where items can be subsequently hidden via hide_ancestor.

tractorcow:
So these have been changed:
- Implemented @chillu's 'FulltextSearch.indexes' config option to globally filter indexes
- Reduced cooldown on the commit task to 5 minutes
- rebased :D

code/search/SearchIndex.php
```diff
@@ -534,10 +542,12 @@ function commit() { }
 
 	public $added = array();
 	public $deleted = array();
+	public $committed = false;
 
 	function reset() {
```
assertchris: Maybe add some visibility keywords... #boyscout

code/search/processors/SearchUpdateCommitJobProcessor.php
```diff
@@ -0,0 +1,247 @@
...
+
+		// Mark complete
+		$this->completed[] = $name;
+	}
+
+	public function setup() {
+		// NOOP
+	}
+
+	public function getJobData() {
```
assertchris: The name of this method is very misleading (especially given the following method's name). Perhaps `getNewJobObject`?

tractorcow: It's an interface method on QueuedJob, so, I can't change it. =(

code/search/SearchIndex.php
tractorcow: Done.

tractorcow:
@chillu is just running a test of this code on 15k documents, so let's see how that turns out. ;D

tractorcow:
I've added a fix to prevent lots of unnecessary duplicate commit jobs being added to the queue.

chillu:
Had a looooong discussion and debugging session with @tractorcow on a large production data set around this pull request. 

The logic could be improved in terms of queueing - it currently starts creating "update" jobs synchronously until the first job actually gets picked up for execution. Since the update job completion triggers the first (and only) "commit" job to be queued, it purely depends on the server's execution speed on how many "update" jobs are in the queue before the first "commit" job. Ideally this would be a fixed interval, depending on the performance of the search warmers (which might be a separate box with altogether separate performance characteristics). But that's for another pull request.

As usual, *legend* Mr. Mooyman!

PR #69, opened 2015-05-07
Regression in SearchUpdater_ObjectHandler

chillu:
Stopped indexing of classes which were unrelated to overall variants.
For example, an index with excludeVariantState(array('SearchVariantVersioned' => 'Stage'))
should only set this variant state on types where appliesTo() returns true, namely "Page".
Without the $class parameter it also returned on "File" index requests,
which then lead to all index requests being discarded later on somewhere in SearchUpdater.

Regression introduced in 625d282.

cc @tractorcow


tractorcow:
Thanks for the fix. :D

PR #70, opened 2015-05-07
NEW Pre-filtering of indexed record through 'list' option in addClass()

chillu:
We needed this as an effective way to filter indexed files. We could've done this through custom properties `(function getIsUsed() {return in_array(...);}` in addition to the more basic `ShowInSearch` flag. But it seemed wasteful to index all files, plus many Solr queries and property setups are more complex than the SS DataQuery API.

Replaces https://github.com/silverstripe-labs/silverstripe-fulltextsearch/pull/25

### DIFF
--- a/code/search/SearchIndex.php
+++ b/code/search/SearchIndex.php
@@ -180,7 +180,9 @@ abstract class SearchIndex extends ViewableData {
 	 *
 	 * @throws Exception
 	 * @param String $class - The class to include
-	 * @param array $options - TODO: Remove
+	 * @param array $options
+	 *  - 'include_children': TODO remove
+	 *  - 'list': A {@link DataList} to pre-filter records to be indexed
 	 */
 	public function addClass($class, $options = array()) {
 		if ($this->fulltextFields || $this->filterFields || $this->sortFields) {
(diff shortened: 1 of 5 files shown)


chillu:
Still working on unit test coverage, wasn't as high as I remembered ;)

hafriedlander:
Hi @chillu. I like the general idea, but a couple of notes:
- I'm not really a fan of `list` as the key, it seems very generic & not very self explanatory. Could we have it be `whenInList` or something?
- Have you thought about performance implications? It's probably OK, but init() function will get run a lot, so something that does a lot of work to calculate `$list` might cause perf issues itself.

tractorcow:
I prefer the approach you originally dismissed, actually. :D

Perhaps look at the way that `addFilterField` currently works, but perhaps by making it act as an explicit pre-filter rather than an optionally filterable field. Rather than specify the explicit list, provide a mechanism (in this case, a field name or getter) to use to check if an item should appear in search? Tell the index "how" to filter, not "what" to filter.

What I'm saying is your getIsUsed() method is much more likely to look like `return $this->ShowInSearch` than `return in_array($key, $listOfIds)`, in practice.

Also, what if you can't determine what the list is at that stage? It might be fine if it's a DataList to evaluate later on down the track, but you'd still need to pass in all the parameters up front, and there's no way to hook into custom getIsUsed logic in the dataobject anyway. It's logic that should be implemented in the model not the index.

dhensby:

robbieaverill:

PR #75, opened 2015-06-26
API Support custom files / spelling correction

tractorcow:
Default config has been fixed (previously would crash on references to name and text fields, when we should use the _text field consistently).

uploadConfig has been refactored out of Solr_Configure into SolrIndex, so that specific indexes can have control over additional files (e.g. synonyms.txt) that are necessary.

Issues with parsing of collation results from solr have been resolved. If the collation result is an object, the suggestions are parsed in turn until a suggestion is returned.

### DIFF
--- a/code/solr/SolrIndex.php
+++ b/code/solr/SolrIndex.php
@@ -141,6 +141,26 @@ abstract class SolrIndex extends SearchIndex {
 		
 		return implode("\n\t\t", $xml);
 	}
+	
+	/**
+	 * Extract first suggestion text from collated values
+	 * 
+	 * @param mixed $collation
+	 * @return string
+	 */
+	protected function getCollatedSuggestion($collation = '') {
+		if(is_string($collation)) {
+			return $collation;
+		}
+		if(is_object($collation)) {
+			if(isset($collation->misspellingsAndCorrections)) {
+				foreach($collation->misspellingsAndCorrections as $key => $value) {
+					return $value;
+				}
+			}
+		}
+		return '';
+	}
 
 	/**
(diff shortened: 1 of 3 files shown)


tractorcow:
This PR has been updated to include a new `_spellcheck` dictionary which is built on the `_spellcheckText` column, if defined. This is useful for sites which may like to build a dictionary on fields other than the default `_text` column.

Also, test failures are pre-existing, but pass locally (with subsites installed)

image

tractorcow:
Rebased with recent fix to unit tests

code/solr/SolrIndex.php
```diff
@@ -472,14 +492,18 @@ function commit() {
...
+	 * @param array $params Extra request parameters passed through to Solr
 	 * @return ArrayData Map with the following keys: 
 	 *  - 'Matches': ArrayList of the matched object instances
 	 */
 	public function search(SearchQuery $query, $offset = -1, $limit = -1, $params = array()) {
 		$service = $this->getService();
 		
-		SearchVariant::with(count($query->classes) == 1 ? $query->classes[0]['class'] : null)->call('alterQuery', $query, $this);
+		SearchVariant::with(
+			count($query->classes) == 1
```
hafriedlander: Can we break this out to a separate statement? I don't like trinaries if we have to break them over so many lines like this.

tractorcow: ok :D

tractorcow:
Restarted timed out tests.

tractorcow:
Thank you @hafriedlander :D

PR #78, opened 2015-06-30
BUG Prevent subsites breaking travis tests

tractorcow:
Fixes https://github.com/silverstripe-labs/silverstripe-fulltextsearch/issues/77

State is not being changed and restored robustly enough for this test to recover, so we simply omit subsites completely from this test.

The problem is likely due to `$illegalExtensions` option in `SapphireTest` not being able to properly restore removed extensions after the test is complete.

It's detrimental to force fulltextsearch tests to rely so heavily on a third party module so until we can fix the feature in core just don't let the tests fail.

### DIFF
--- a/tests/BatchedProcessorTest.php
+++ b/tests/BatchedProcessorTest.php
@@ -48,15 +48,27 @@ class BatchedProcessorTest extends SapphireTest {
 		)
 	);
 
+	public function setUpOnce() {
+		// Disable illegal extensions if skipping this test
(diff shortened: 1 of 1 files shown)


tractorcow:
Green, please merge @hafriedlander :D

tractorcow:
Another approach; Disabling the "hazardous" test only if subsites is installed.

tractorcow:
merge before it breaks again!

tractorcow:
Thanks @hafriedlander :D

PR #80, opened 2015-07-06
Added SynonymValidator

assertchris:
Cinnamon validator. Hmmm...

### DIFF
--- a/tests/SynonymValidateTest.php
+++ b/tests/SynonymValidateTest.php
@@ -1,0 +1,84 @@
+<?php
+
+class SynonymValidatorTest extends PHPUnit_Framework_TestCase {
+	/**
+	 * @var SynonymValidator
+	 */
+	protected $validator;
+
+	/**
+	 * @inheritdoc
+	 */
+	public function setUp() {
+		parent::setUp();
+
+		$this->validator = new SynonymValidator(array(
+			'Synonyms',
+		));
+	}
+
+	/**
+	 * @inheritdoc
+	 */
+	public function tearDown() {
+		$this->validator = null;
+
+		parent::tearDown();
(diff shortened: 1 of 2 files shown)


tractorcow:
This looks good, but maybe we should put this in the CWP module instead of the open source one? Without the cwp core api this doesn't actually have any use case.

assertchris:
I dunno. It does seem reasonable to see this if non-CWP applications want CMS-controlled synonyms...

tests/SynonymValidateTest.php
```diff
@@ -0,0 +1,84 @@
...
+			array('foo, bar ,baz'),
+			array('
+				foo
+				bar
+				baz
+			'),
+			array('
+				# this is a comment, it should be ignored!
+
+				foo=>bar,baz
```
tractorcow: I think we need to support `foo => bar, baz`. Also we can only have one `=>` per line :D

tractorcow:
There's no CMS control for synonyms in this module though.

assertchris:
I see that. The feature is tied to enforcing valid Solr configuration values, and can be used without being linked to a form (as the unit tests demonstrate). I don't mind where it goes, but I see merit in it living in either place.

tests/SynonymValidateTest.php
assertchris: ok!

tractorcow:
Can you put it with the SynonymsSiteConfig class in cwp? You'll need a PR to that module anyway in order to ensure the validator is enabled. :D

assertchris:
Ok, moved to https://gitlab.cwp.govt.nz/cwp/cwp/merge_requests/82.

PR #82, opened 2015-07-23
API QueuedJob support for Solr_Reindex

tractorcow:
This also has incremental deletion of records; Rather than deleting all records in solr up front, it'll simply delete all records that it KNOWS won't exist up front, and then deletes all other records incrementally in batches.

Each batch is derived using the formula `ID % total_groups = this_group`. This means that batches don't need to store a list of IDs for each record, and copes with cases where ids may not be contiguous, as well as deleted IDs that may not have a local record anymore.

Re-index is broken into a base implementation, with separate subclasses for each implementation (including immediate reindex for local, message queue and queuedjobs).

 `Solr_Reindex`, two queued jobs, a new logging API, and several tests have been added.

Part of the logic for generating SOLR query syntax has been moved out of SolrIndex and into SearchQuery. This is so that SearchQuery may be used to generate syntax for deletion of records, which is necessary for the incremental deletion feature.

### DIFF
--- a/code/solr/reindex/handlers/SolrReindexImmediateHandler.php
+++ b/code/solr/reindex/handlers/SolrReindexImmediateHandler.php
@@ -1,0 +1,74 @@
+<?php
+
+use Psr\Log\LoggerInterface;
+
+/**
+ * Invokes an immediate reindex
+ *
+ * Internally batches of records will be invoked via shell tasks in the background
+ */
+class SolrReindexImmediateHandler extends SolrReindexBase {
+
+	public function triggerReindex(LoggerInterface $logger, $batchSize, $taskName, $classes = null) {
+		$this->runReindex($logger, $batchSize, $taskName, $classes);
+	}
+
+	protected function processIndex(
+		LoggerInterface $logger, SolrIndex $indexInstance, $batchSize, $taskName, $classes = null
+	) {
+		parent::processIndex($logger, $indexInstance, $batchSize, $taskName, $classes);
+
+		// Immediate processor needs to immediately commit after each index
+		$indexInstance->getService()->commit();
+	}
+
+	/**
+	 * Process a single group.
+	 *
+	 * Without queuedjobs, it's necessary to shell this out to a background task as this is
+	 * very memory intensive.
+	 *
+	 * The sub-process will then invoke $processor->runGroup() in {@see Solr_Reindex::doReindex}
+	 *
+	 * @param LoggerInterface $logger
+	 * @param SolrIndex $indexInstance Index instance
+	 * @param array $state Variant state
+	 * @param string $class Class to index
+	 * @param int $groups Total groups
+	 * @param int $group Index of group to process
+	 * @param string $taskName Name of task script to run
+	 */
+	protected function processGroup(
+		LoggerInterface $logger, SolrIndex $indexInstance, $state, $class, $groups, $group, $taskName
+	) {
+		// Build state
+		$statevar = json_encode($state);
+		if (strpos(PHP_OS, "WIN") !== false) {
+			$statevar = '"'.str_replace('"', '\\"', $statevar).'"';
+		} else {
+			$statevar = "'".$statevar."'";
+		}
+
+		// Build script
+		$indexName = $indexInstance->getIndexName();
+		$scriptPath = sprintf("%s%sframework%scli-script.php", BASE_PATH, DIRECTORY_SEPARATOR, DIRECTORY_SEPARATOR);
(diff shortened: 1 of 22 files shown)


sminnee:
OK I've done a bunch of comments on particular lines. Most of them relate to how logging is handled. I'd be keen to get @assertchris' views on it.

I think that it would be okay to introduce some new API features into a minor release of queuedjobs and depend on that, rather than increase the overall complexity just to avoid requiring queuedjobs changes.

assertchris:
We don't have to require `psr/log`, just a compliant logger (like `monolog`) to be PSR compliant. It's definitely an easy thing to do in this case, and can set a president for rolling out this PSR in the rest of the codebase (to replace SS_Logger). Or we can create a proxy which has SS_Logger API sent to Monolog. I've started and stopped this multiple times already...

sminnee:
I figured that referencing psr/log would mean that we can couple to the interface, not the implementation. If we depended on monolog then it would be harder to swap out the logging implementation.

Presumably the user of queuedjobs would need to include and configure monolog and then inject it into this. In SS3 this would be a bit manual but we can add some stuff in SS4 to make it a bit more straightforward?

assertchris:
Yeah, but [Monolog requires `psr/log`](https://github.com/Seldaek/monolog/blob/e4f45be5dcfeb54c45f85b50cdbb3ef8a2bfbd81/composer.json#L17) already...

sminnee:
I guess if we were coupling to a specific implementation anyway ( eg for its null logger ) that would be okay, but given the choice I don't see why queuedjobs should need to be away of monolog' existence.

In other news, this issue would help tidy up the use of PSR-3: https://github.com/silverstripe/silverstripe-framework/issues/4425

assertchris:
Yeah, you're right.

tractorcow:
composer-level dependency injection? Nevermind I sense a can of worms going down that path. :D

tractorcow:
Hi @sminnee please have a look at the logging replacement at https://github.com/tractorcow/silverstripe-fulltextsearch/commit/d661696a473fb50ec322e033e0c770cdbce948c3

sminnee:
Cool - this looks much better (except for the class misname you identified). Squash before merging?

The code setting up the logs in Solr_BuildTask would ideally be part of the framework, but that's probably more a SS4 thing, and can wait.

The only thing I wonder is whether it would be best to put QueueJobLogHandler into the queuedjobs module. It could be left for a future refactor, but refactoring a class across a module is a bit annoying... If you were to do that, I would include `getLogger()` in the base class for all queued jobs, and say "Returns a LoggerInterface that you can use to supply messages for the job" or something...

tractorcow:
@sminnee I've moved the setup into a factory, so anything monolog-specific is abstracted into one file.

Thanks to @assertchris for the kind help and advice on this. :D

tractorcow:
I think the best place to start is in core / framework and implement logging properly there; Then we can make sure that both queuedjobs and fulltextsearch use the same solution.

tractorcow:
Fixed to remove solr-specific stuff from SearchQuery.php

tractorcow:
I've updated this PR to allow cancelling of in-progress tasks.

See https://github.com/open-sausages/silverstripe-queuedjobs/pull/6 for necessary queuedjobs pr

tractorcow:
This code should work fine with existing queuedjobs, but has been updated to work best with the new doorman engine.

See https://github.com/silverstripe-australia/silverstripe-queuedjobs/pull/47

code/solr/reindex/handlers/SolrReindexImmediateHandler.php
```diff
@@ -0,0 +1,74 @@
...
+		if (strpos(PHP_OS, "WIN") !== false) {
+			$statevar = '"'.str_replace('"', '\\"', $statevar).'"';
+		} else {
+			$statevar = "'".$statevar."'";
+		}
+
+		// Build script
+		$indexName = $indexInstance->getIndexName();
+		$scriptPath = sprintf("%s%sframework%scli-script.php", BASE_PATH, DIRECTORY_SEPARATOR, DIRECTORY_SEPARATOR);
+		$scriptTask = "php {$scriptPath} dev/tasks/{$taskName}";
```
assertchris: Is it safe to assume `php` will be found? perhaps `PHP_BIN . "/php"`?

tractorcow: This is what the code was using before, so I guess this just keeps the same resolution mechanism in place.

Although in the past the php reference was in Solr_Reindex class.

PR #85, opened 2015-10-01
Added hooks to allow existing Solr run methods to be extended

textagroup:
Adds a hook in the run methods of Solr_BuildTask, Solr_Configure and Solr_Reindex.

My personal reason for adding this is for better namespace support as currently a HTTP request is used to store the classes to be handled and when using namespaces this currently gets sanitised.

### DIFF
--- a/code/solr/Solr.php
+++ b/code/solr/Solr.php
@@ -195,6 +195,7 @@ class Solr_BuildTask extends BuildTask {
 	 * @param SS_HTTPReqest $request
 	 */
 	public function run($request) {
+		$this->extend('updateRun', $request);
 		$name = get_class($this);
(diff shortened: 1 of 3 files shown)


dhensby:
You should only need the extend call once in the base class.

dhensby:
We'll need to better support namespaces with SS4 so this change will be valuable.

I'm not a fan of the phockito change being part of this PR. @textagroup can I ask you to amend this PR so it only contains the commit regarding the extensions and you place ONE extension point in the base function, please?

robbieaverill:
This change has been superceded by #155 which does the same thing as this but with a different hook name, thanks for contributing @textagroup - closing for now

PR #86, opened 2015-10-06
MINOR: Don't include Hamcrest globally so it doesn't conflict with PHPUnit

madmatt:
PHPUnit now includes matching functions like `any()` and `anything()` outside of a namespace, which conflicts with Phockito that does the same thing when calling `Phockito::include_hamcrest(true)`.

This test in particular breaks Behat, because the file includes a class (`SolrReindexTest_Item`) that extends `DataObject`, meaning that this file gets included very early in the initialisation process.

Instead, we can just set `Phockito::include_hamcrest(false)` and use the full path to the Hamcrest matching functions instead (e.g. `\Hamcrest_Matchers::anything()`).

### DIFF
--- a/tests/SolrReindexTest.php
+++ b/tests/SolrReindexTest.php
@@ -5,7 +5,7 @@ use Monolog\Handler\HandlerInterface;
 use Monolog\Logger;
 use Psr\Log\LoggerInterface;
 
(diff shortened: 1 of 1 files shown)


madmatt:
I'm closing this now and opening a new PR instead, which updates everything instead of just the one broken test.

PR #87, opened 2015-10-08
adding ability for solr suffix and prefixing

pitchandtone:
This means the same code can run on a single solr core and have the same indexes but use prefixes or suffixes to keep them separate

### DIFF
--- a/code/solr/SolrIndex.php
+++ b/code/solr/SolrIndex.php
@@ -80,9 +80,25 @@ abstract class SolrIndex extends SearchIndex
         return $this->renderWith($this->getTemplatesPath() . '/schema.ss');
     }
 
+    /**
+     * Helper for returning the correct index name. Supports prefixing and
+     * suffixing
+     *
+     * @return string
+     */
     public function getIndexName()
     {
-        return get_class($this);
+        $name = get_class($this);
+
+        if(defined('SS_SOLR_INDEX_PREFIX')) {
+            $name = SS_SOLR_INDEX_PREFIX . ''. $name;
+        }
+
+        if(defined('SS_SOLR_INDEX_SUFFIX')) {
+            $name = $name . '' . SS_SOLR_INDEX_SUFFIX;
+        }
+
+        return $name;
     }
 
     public function getTypes()
--- a/code/solr/reindex/handlers/SolrReindexImmediateHandler.php
+++ b/code/solr/reindex/handlers/SolrReindexImmediateHandler.php
@@ -52,9 +52,10 @@ class SolrReindexImmediateHandler extends SolrReindexBase
 
         // Build script
         $indexName = $indexInstance->getIndexName();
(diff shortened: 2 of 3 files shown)


normann:
@hafriedlander / anyone who is active on this module, this one is exactly for a situation that needs to create two cores on one host server, using one site (code), could you guys have a look, please? 
Though I think the pull request's description is a bit blur. misleading.

tractorcow:
I'm concerned that there could be other parts of the code that rely on the assumption that indexName === indexClass. It would require a new major version to change this assumption.

Could we, instead of this change, put the new code into a new `public function getIndexStoredName()` and subsitute that where the index name is otherwise passed to solr?

getIndexName() could remain the original definition (and still be used to lookup the class).

In any case, we probably need to do an exhaustive search of the API to ensure the right value is used in each case.

normann:
@tractorcow either ways should work.

JFI, this pull request works in these situations or combinations of these setting:

Environment set as 'dev', so 'SearchUpdateImmediateProcessor' / 'SolrReindexImmediateHandler' used, no matter what.
Environment set as 'test' and QueuedJob module present, so 'SearchUpdateQueuedJobProcessor' and 'SolrReindexQueuedHandler' used.
Have Checked both Configure/Reindex initial process and triggered some process to individually update objects that previously indexed, all works.

Though, I didn't check the situation of using 'SearchUpdateMessageQueueProcessor' and 'SolrReindexMessageHandler'.

I am not objected to the idea to separate the function into 2 functions and use them semantically, that seemed a cleaner way.

tractorcow:
Ok, that gives me a bit more confidence in this PR. :)

Do you want to rebase so we can do some more testing?

tractorcow:
If @pitchandtone isn't available we might need to rebase it for him in another PR....

pitchandtone:
Yeah sorry about that. Won't be back in action until next week...

tractorcow:
All good; Sorry to let this sit for so long. ;)

pitchandtone:
Travis failure seem unrelated.

dhensby:
travis is broken on this repo atm

tractorcow:
Ugh, was that that versioning test I was ment to fix before 3.2.3? I guess too late now...

dhensby:
Yep...

tractorcow:
I'd like @hafriedlander or @chillu to look at this before merging. I'm not confident that this won't have regressions.

dhensby:
bump @chillu @hafriedlander - I can confirm I'm using this patch in production

pitchandtone:
@dhensby @hafriedlander works fine on prod for Wellingtonnz.com

dhensby:
really sorry @pitchandtone - this needs a rebase now, then we'll get it in

pitchandtone:
@dhensby rebased now.

dhensby:
Thanks

PR #109, opened 2016-02-23
FIX Indexes with custom index names that dont match the classname wer…

dhensby:
…e breaking

---

At the moment if you define an index like so:

```php

class MyIndex extends SolrIndex {

    public function init() {
        ...
    }

    public function getIndexName() {
        return "some-string";
    }

}
````

Then on Reindex task the following error is output:

```
ERROR [User Error]: Uncaught ReflectionException: Class some-string does not exist
IN GET /dev/tasks/Solr_Reindex
Line 14 in /vagrant/www/car.local/public_html/framework/control/injector/InjectionCreator.php
```

This patch fixes that by searching for the SolrIndex by IndexName if the name does not exist as a class.

### DIFF
--- a/code/solr/Solr.php
+++ b/code/solr/Solr.php
@@ -107,8 +107,8 @@ class Solr
 
     /**
      * Get a SolrService
-     * 
-     * @param string $core Optional core name
+     *
+     * @param string $core Optional name of index class
      * @return SolrService_Core
      */
     public static function service($core = null)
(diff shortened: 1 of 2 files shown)


dhensby:
@chillu @tractorcow ;)

tractorcow:
I promised to look at the versioned test... it's due to a regression in the fix I made for versioned in framework. Might need to revert it... but not urgent.

dhensby:
I'm going to remove my last commit messing with the `.travis.yml`. The errors I'm getting are happening on current master too (see restarted test here: https://travis-ci.org/silverstripe-labs/silverstripe-fulltextsearch/jobs/106883289)

tractorcow:
I'll try to fix these issues before we release the next 3.2

dhensby:
Any chance of a merge given the test failures are unrelated?

dhensby:
thanks

tractorcow:
Promised fix at https://github.com/silverstripe-labs/silverstripe-fulltextsearch/pull/114

PR #113, opened 2016-04-12
BUGFIX: Fixed issue where the $id variable would be overridden in derived fields

UndefinedOffset:
When using derived fields subsequent iterations of the loop on [line 527](https://github.com/silverstripe-labs/silverstripe-fulltextsearch/blob/master/code/search/SearchIndex.php#L527) can in some cases be reset to an id that is not the one being re-indexed resulting in a SQL error.

### DIFF
--- a/code/search/SearchIndex.php
+++ b/code/search/SearchIndex.php
@@ -552,8 +552,8 @@ abstract class SearchIndex extends ViewableData
                         $dirty[$base] = array();
                     }
 
-                    foreach ($ids as $id) {
-                        $statefulid = array('id' => $id, 'state' => $state);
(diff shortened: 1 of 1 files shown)


UndefinedOffset:
Looks like the travis build is failing because of an unrelated issue see previous build: https://travis-ci.org/silverstripe-labs/silverstripe-fulltextsearch/builds/117662029

tractorcow:
Please rebase on master; https://github.com/silverstripe-labs/silverstripe-fulltextsearch/pull/114 has fixed the broken build.

UndefinedOffset:
@tractorcow looks like its passing now :)

PR #114, opened 2016-04-15
BUG Fix versioned writes where subtables have no fields key

tractorcow:
Fixes some regressions in recent framework fixes to versioned.

Also remove unnecessary config nesting in tests which are now handled via core.

Fixes issue I mentioned at https://github.com/silverstripe-labs/silverstripe-fulltextsearch/pull/109#issuecomment-189010025

cc @dhensby , sorry it's late!

### DIFF
--- a/code/search/SearchUpdater.php
+++ b/code/search/SearchUpdater.php
@@ -90,14 +90,14 @@ class SearchUpdater extends Object
         $writes = array();
 
         foreach ($manipulation as $table => $details) {
-            if (!isset($details['id']) || !isset($details['fields'])) {
+            if (!isset($details['id'])) {
                 continue;
             }
 
             $id = $details['id'];
             $state = $details['state'];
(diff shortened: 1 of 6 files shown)


code/search/SearchUpdater.php
```diff
@@ -90,14 +90,14 @@ public static function handle_manipulation($manipulation)
         $writes = array();
 
         foreach ($manipulation as $table => $details) {
-            if (!isset($details['id']) || !isset($details['fields'])) {
+            if (!isset($details['id'])) {
```
tractorcow: if `fields` is empty, we still need to run this loop to trigger the `update to more specific classname` line below.

dhensby:
green ticks, yay!

dhensby:
Let's tag a release?

tractorcow:
Soon; I need to merge a variants fix first.

PR #116, opened 2016-04-21
FIX: Fix initial dev/build on PDO Database.

sminnee:
When the database isn’t yet created, $current is set, but
$current->currentDatabase() is empty. I suspect this only applies to 
PDO connections. It results in errors during startup.

This check fixes it.

### DIFF
--- a/code/search/SearchUpdater.php
+++ b/code/search/SearchUpdater.php
@@ -22,7 +22,7 @@ class SearchUpdater extends Object
         global $databaseConfig;
 
         $current = DB::getConn();
-        if (!$current || @$current->isManipulationCapture) {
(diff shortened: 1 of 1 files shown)


tractorcow:
@sminnee can you show me the errors during setup? If we skip bind here, does it re-bind later once the connection selects a database?

sminnee:
It only occurs before the database first exists. The error was that it was trying to call `USE ""` on the database in the subsequent `$captured->selectDatabase($current->currentDatabase())`.

It'll rebind on any subsequent pageviews.

tractorcow:
ok, sounds good to me. If you rebase I'll merge, thank you. :D

PR #117, opened 2016-04-21
BUG fix issues with search variants applying to more than one class

tractorcow:
Plus lots of other fixes:
- Fixed more than one versioned object in an index
- Fixed more than one subsites object in an index
- Fixed SearchVariantSubsites ending up with unflushable (and often invalid) caches of subsites during write
- Fix empty catch blocks from hiding errors during testing
- Add missing abstract alterDefinition to SearchVariant.
- Fix variants assuming that origin === base class for each object type added

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,36 +1,35 @@
 # See https://github.com/silverstripe-labs/silverstripe-travis-support for setup details
 
-sudo: false
-
 language: php
 
+sudo: false
+
 php:
-  - 5.3
   - 5.4
   - 5.5
   - 5.6
-  - 7.0
 
 env:
   - DB=MYSQL CORE_RELEASE=3.2
 
 matrix:
   include:
+    - php: 5.3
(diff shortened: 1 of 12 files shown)


tractorcow:
I've put back the subsites config in .travis.yml. Hopefully it'll now run tests with subsites. :)

tractorcow:
These tests are a bunch of garbage.

tractorcow:
Fixed with https://github.com/silverstripe/silverstripe-cms/pull/1457

tractorcow:
green

dhensby:
Can this go into master atm? Master is the 2.1.x-dev branch, so should only take patches and I'm seeing a BC breakage by adding `abstract public function alterDefinition($class, $index);`

tractorcow:
I'll remove it.

tractorcow:
alterDefinition is now removed

PR #118, opened 2016-05-08
Add an extension hook to allow altering search results

DarrenInwood:
Often I find myself copying the entire search() function from SolrIndex into my custom index classes, so that I can pull some extra information from the solr-php-client into the templates.

I've included some docs on how to use this extension point to add faceting, a common use case.

This change didn't seem to need tests.

### DIFF
--- a/code/solr/SolrIndex.php
+++ b/code/solr/SolrIndex.php
@@ -795,7 +795,13 @@ abstract class SolrIndex extends SearchIndex
             }
         }
 
-        return new ArrayData($ret);
+        $ret = new ArrayData($ret);
+
+        // Enable extensions to add extra data from the response into
+        // the returned results set.
+        $this->extend('updateSearchResults', $ret, $res);
+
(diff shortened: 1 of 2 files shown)


dhensby:
Looks good. @tractorcow we need to adjust our branches. master is a patch branch according to the composer alias so new features can't go there :(

tractorcow:
No, it's ok to put new features into master.

When we tag I'll do a semver review to check the version I should tag. It's an "at realease" job not an "at merge" one. :D

dhensby:
If our master is aliased as Y.Z.x-dev it would be reasonable for devs using that in computer to expect it only receives patches

tractorcow:
Then, um... I guess maybe we should do branching at-merge in that case.

dhensby:
I'd always make master N.x-dev where N is 1 greater than the latest major release otherwise we do have a problem at merge time for people doing breaking changes and what not

PR #119, opened 2016-05-24
BUGFIX: Fixed crash when the relationship has no parents

UndefinedOffset:
When using derived fields if the ids returned from the previous iteration of the [chain loop](https://github.com/silverstripe-labs/silverstripe-fulltextsearch/blob/master/code/search/SearchIndex.php#L547-L560) is empty an SQL error will occur due to the ``IN()`` statement being empty. This pull request simply breaks out of the loop when ``$ids`` is empty.


dhensby:
Thanks

PR #121, opened 2016-06-13
API Moving index finding to a static function

dhensby:
For a 2.3 release

Moving this out to it's own function so other modules can use it

### DIFF
--- a/code/solr/Solr.php
+++ b/code/solr/Solr.php
@@ -140,6 +140,38 @@ class Solr
     }
 
     /**
+     * Find an index class from the index name - this is for when index names are not the name of the class
+     *
+     * @param string $indexName The index name to search for
+     * @return string
+     */
+    public static function get_index_from_name($indexName) {
+        $index = $indexName;
+
+        //find the index class by IndexName
+        // this is for when index names do not match the class name
+        // (this can be done by overloading getIndexName() on indexes
+        if ($indexName && !ClassInfo::exists($index)) {
+            foreach(self::get_indexes() as $solrIndexClass) {
+                $reflection = new ReflectionClass($solrIndexClass);
+                //skip over abstract classes
+                if (!$reflection->isInstantiable()) {
+                    continue;
+                }
+                //check the indexname matches the index passed to the request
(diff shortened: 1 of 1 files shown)


code/solr/Solr.php
```diff
@@ -140,6 +140,38 @@ public static function get_indexes()
...
+            foreach(self::get_indexes() as $solrIndexClass) {
+                $reflection = new ReflectionClass($solrIndexClass);
+                //skip over abstract classes
+                if (!$reflection->isInstantiable()) {
+                    continue;
+                }
+                //check the indexname matches the index passed to the request
+                if (!strcasecmp(singleton($solrIndexClass)->getIndexName(), $indexName)) {
+                    //if we match, set the correct index name and move on
+                    $index = $solrIndexClass;
```
tractorcow: Just return here?

code/solr/Solr.php
```diff
@@ -140,6 +140,38 @@ public static function get_indexes()
...
+                    break;
+                }
+            }
+        }
+
+        return $index;
```
tractorcow: Where's the error where an invalid $index is passed in? It shouldn't return $index if that's not a valid name.

code/solr/Solr.php
```diff
@@ -140,6 +140,38 @@ public static function get_indexes()
...
+        $index = $indexName;
+
+        //find the index class by IndexName
+        // this is for when index names do not match the class name
+        // (this can be done by overloading getIndexName() on indexes
+        if ($indexName && !ClassInfo::exists($index)) {
+            foreach(self::get_indexes() as $solrIndexClass) {
+                $reflection = new ReflectionClass($solrIndexClass);
+                //skip over abstract classes
+                if (!$reflection->isInstantiable()) {
```
tractorcow: FullTextSearch::get_indexes() already filters these.

robbieaverill:
@dhensby would you like this to be revived again?

robbieaverill:
Feel free to reopen if you'd like to do this in a newer release line =)

PR #122, opened 2016-06-16
BUGFIX: SolrVersionedTest_Index causing crash in unit tests when Phockito is not installed

UndefinedOffset:
When Phockito is not installed the SolrVersionedTest_Index causes an invalid argument exception to be thrown. Stating "Can't add classes which don\'t have data tables (no $db or $has_one set on the class)". This appears to be caused by the return in SolrIndexVersionedTest.php inside of a check for the Phockito class. Moving the SolrIndexVersionedTest_Object class above that check stops the exception. This pull request does that.


tractorcow:
Thanks!

PR #128, opened 2016-09-21
Remove Porter Stemming from default template

brettt89:
Porter Stemming is causing synonyms to respond incorrectly and provide skewed results. Removing Porter Stemming from default template allows functionality to work out of the box.

Stemming can be added by client by replacing type.ss template if required.


dhensby:
Thanks

PR #131, opened 2016-12-02
FIX Regression where pages would be indexed under all subsites

dhensby:


code/search/SearchVariantSubsites.php
```diff
@@ -83,9 +83,16 @@ public function extractManipulationWriteState(&$writes)
                 continue;
             }
 
+            if (isset($write['fields']['SiteTree:SubsiteID'])) {
```
tractorcow: Can we add `'File:SubsiteID'` as well?

dhensby: Files in subsite 0 should be global, but I suppose that means they should still only be indexed once.... So... Yes?

tractorcow: Yes but files in subsiteid = 3 shouldn't be.

If they are global, they must be indexed in EVERY subsite, not only once.

dhensby: :(

tractorcow: Unfortunately Solr acts as an extremely de-normalised store; For any filter (e.g. any given subsite, stage, etc combination) a filter will be created; This is that json-looking string in the indexed data. If a particular model is visible within that filter (such as golbal files) then there needs to be a separate record saved for that combination.

dhensby:
OK - added the `File:SubsiteID` check too

tractorcow:
It looks good, but should we have a test too?

dhensby:
I've added tests, but because the travis fix to find best matching tag/branch isn't in the CI suite isn't running

tractorcow:
I've merged the travis fix; Restarting tests. :)

tractorcow:
See https://github.com/silverstripe/silverstripe-travis-support/pull/38#issuecomment-265860021

tractorcow:
Thats some fine work @dhensby :)

dhensby:
oh darn it, this was meant to be in 2.2!

PR #137, opened 2017-03-07
Fix #134 - archiving and unpublishing now removes Live variant of ver…

### LINKED ISSUE #134: Archived and Unpublished pages do not remove 'Live' variant from index
To replicate:

Create a new page and save it as draft.
Search for the page (in incognito mode) - 0 results.

Publish the page.
Search for the page - 1 result.

Unpublish the page.
Search for the page.
Expected - 0 results.
Actual - PaginatedList reports total size of 1, but is empty.

Publish the page, then Archive it.
Search for the page.
Expected - 0 results.
Actual - PaginatedList reports total size of 1, but is empty.

Versioned alters this query so that we only get the item if it has a live version; the item is being returned from the index still, but is not being added to the list.
https://github.com/silverstripe/silverstripe-fulltextsearch/blob/master/code/solr/SolrIndex.php#L749

However, the total including the ones filtered out is being set as the PaginatedList total:
https://github.com/silverstripe/silverstripe-fulltextsearch/blob/master/code/solr/SolrIndex.php#L786

As a workaround I add these lines to affected DataObjects.

`
    public function onBeforeDelete()
    {
        $this->LastEdited = time();
        $this->write();
    }

    public function onAfterUnpublish()
    {
        singleton('nameofyourindex')->delete(
            ClassInfo::baseDataClass($this->ClassName),
            $this->ID,
            array (
              'SearchVariantVersioned' => 'Live',
              'SearchVariantSubsites' => '0',
            )
        );
    }
`

DarrenInwood:
…sioned items from index.

This ensures that all Versioned variants (ie. Stage and Live) are checked and deleted if necessary on every write.

Previously, using the CMS functions for Archive or Unpublish would result in a stale Live variant remaining in the index.  This would be returned in search results from Solr, but would fail to be added into the PaginatedList returned by the search function, resulting in eg. 'Showing 1-10 of 22 results' but actually there are only 8 visible since 2 have been deleted and don't exist in the database.

I couldn't decipher the test suite in the time I had available to add automated tests for this, sorry, but have tested manually using the following procedure:

* Perform CMS operation
* Find what's in the index after the operation - http://<server>:8983/solr/<index>/select?q=*%3A*&fl=_documentid,SiteTree_Title&wt=json&indent=true

Tested operations:
* Save first draft
* Publish
* Save new draft
* Cancel draft changes
* Revert to previous version (no published version)
* Unpublish (no later draft)
* Unpublish (later draft exists)
* Restore draft version
* Archive (No published version)
* Archive (Published version)

### DIFF
--- a/code/search/SearchVariantVersioned.php
+++ b/code/search/SearchVariantVersioned.php
@@ -66,6 +66,36 @@ class SearchVariantVersioned extends SearchVariant
         }
     }
 
+    /**
+     * If we are doing a delete or an unpublish where the *_Live table had a
+     * delete operation, this isn't included in the database manipulation array.
+     *
+     * We need to ensure each state is checked on every write.
+     */
+    public function extractManipulationWriteState(&$writes)
+    {
+        $self = get_class($this);
+
+        foreach ($writes as $key => $write) {
+            $applies = $this->appliesTo($write['class'], true);
+            if (!$applies) {
+                continue;
+            }
+
+            $reindexStates = $this->reindexStates();
+            $next = array();
+            foreach ($write['statefulids'] as $i => $statefulid) {
+                // Add copies of the state to write with all possible Versioned states
+                foreach ($reindexStates as $reindexState) {
+                    $reindexStatefulid = $statefulid;
+                    $reindexStatefulid['state'][$self] = $reindexState;
+                    $next[] = $reindexStatefulid;
+                }
+            }
+            $writes[$key]['statefulids'] = $next;
+        }
+    }
+
     public function extractStates(&$table, &$ids, &$fields)
     {
         $class = $table;
--- a/tests/SearchVariantVersionedTest.php
+++ b/tests/SearchVariantVersionedTest.php
@@ -53,11 +53,14 @@ class SearchVariantVersionedTest extends SapphireTest
         $item->publish("Stage", "Live");
 
         SearchUpdater::flush_dirty_indexes();
-        $this->assertEquals(self::$index->getAdded(array('ID', '_versionedstage')), array(
+        // Note: All states are checked on each action
+        $expectedStates = array(
+            array('ID' => $item->ID, '_versionedstage' => 'Stage'),
             array('ID' => $item->ID, '_versionedstage' => 'Live')
-        ));
+        );
+        $this->assertEquals($expectedStates, self::$index->getAdded(array('ID', '_versionedstage')));
 
-        // Just update a SiteTree field, and check it updates Stage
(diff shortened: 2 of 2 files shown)


tractorcow:
@DarrenInwood I've reviewed the code and it seems that it's correct, but tests need to be updated.

Can you please add the additional assertion that the `$item->publish("Stage", "Live");` in `SearchVariantVersionedTest` writes both stage and live records? Previously it would only write live, but now it should write stage as well (it seems).

Can you please also add a `$item->deleteFromStage('Live')` just to make sure that it's removed from live stage? That would sufficiently provide 100% test coverage for the fixed issue. :D Thank you.

tractorcow:
Solution DOES work on doing a write. It does NOT work if doing a delete only.

E.g. `$item->deleteFromStage('Live');` without a subsequent write will do nothing.

DarrenInwood:
Can you give me a hint on how to run tests for the module?

I get:

    root@0d88caae43bf:/var/www/html# sudo -u www-data vendor/bin/phpunit fulltextsearch/tests
    Fatal error: Class 'SiteTree' not found in /var/www/html/fulltextsearch/tests/BatchedProcessorTest.php on line 4

and:

    root@0d88caae43bf:/var/www/html# sudo -u www-data php framework/cli-script.php dev/tests/SolrVariantVersionedTest flush=1
    ERROR [User Error]: TestRunner::only(): Invalid TestCase 'SolrVariantVersionedTest', cannot find matching class
    IN GET /dev/tests/SolrVariantVersionedTest
    Line 235 in /var/www/html/framework/dev/TestRunner.php

I can run tests against mysite/tests no problem.

tractorcow:
Hm, the first one should work. Do you have CMS installed too?

tractorcow:
I changed my mind... I think we should close this fix in favour of updating https://github.com/silverstripe/silverstripe-fulltextsearch/pull/46. How do you feel about this @DarrenInwood ?

DarrenInwood:
I didn't see #46 - as long as it solves the problem, it looks like they are roughly equivalent.  #46 is easier to understand what's happening at a glance, and #137 is overkill in that it blindly updates both variant states even when not needed, so yep go with #46.

The conclusion I came to in the end is that Versioned should keep some sort of event system or onAfterX/onBeforeX hook system separate from the database-level manipulations, and the SearchVariant should tap into that instead.

I'll leave this for you to close up if you decide to go that way.

tractorcow:
Replaced with https://github.com/silverstripe/silverstripe-fulltextsearch/pull/138

PR #138, opened 2017-03-08
BUG Fix delete / unpublish

### LINKED ISSUE #134: Archived and Unpublished pages do not remove 'Live' variant from index

tractorcow:
Split SearchUpdater classes into separate files
Fixes #134

### DIFF
--- a/tests/SearchVariantVersionedTest.php
+++ b/tests/SearchVariantVersionedTest.php
@@ -72,6 +72,28 @@ class SearchVariantVersionedTest extends SapphireTest
         ));
         $added = self::$index->getAdded(array('ID', '_versionedstage'));
         $this->assertEquals($expected, $added);
+
+        // Test unpublish
+
+        self::$index->reset();
+
+        $item->deleteFromStage('Live');
+
+        SearchUpdater::flush_dirty_indexes();
+
+        $this->assertEquals(1, count(self::$index->deleted));
+        $this->assertEquals(
+            'SiteTree',
+            self::$index->deleted[0]['base']
+        );
+        $this->assertEquals(
+            $item->ID,
+            self::$index->deleted[0]['id']
+        );
+        $this->assertEquals(
+            'Live',
+            self::$index->deleted[0]['state']['SearchVariantVersioned']
+        );
     }
 
     public function testExcludeVariantState()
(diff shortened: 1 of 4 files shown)


tractorcow:
This works, where the existing solution didn't because:
 - Versioned nukes any manipulation against a class without a table. This means if the class only had fields in a parent class (E.g. Page, occasionally) the deletion wouldn't work. Fixed by using the entire hierarchy, so at least one table gets passed through.
 - SearchUpdater nukes any manipulation with empty fields key. Fixed by adding an special 'delete' action and having the updater exclude that from the filter.

tractorcow:
Green. :D todo,merge into master once merge into 2.3

tests/SearchVariantVersionedTest.php
```diff
@@ -72,6 +72,28 @@ public function testPublishing()
...
+
+        // Test unpublish
+
+        self::$index->reset();
+
+        $item->deleteFromStage('Live');
+
+        SearchUpdater::flush_dirty_indexes();
+
+        $this->assertEquals(1, count(self::$index->deleted));
```
dhensby: `$this->assertCount(1, self::$index->deleted)`

dhensby:
merged into master

PR #139, opened 2017-03-28
Update Solr.md

silverstripe-elliot:
Added documentation for how to search on macrons and other unicode characters. Thanks to @chrispenny for the code sample.

### DIFF
--- a/docs/en/Solr.md
+++ b/docs/en/Solr.md
@@ -192,6 +192,20 @@ from a new file `mysite/solr/templates/types.ss` instead:
 		}
 	}
 
+#### Searching for macrons and other Unicode characters
+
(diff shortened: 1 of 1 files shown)


dhensby:
should this not be on by default?

chrispenny:
I'm not sure if you're question is asking for this setting to be added as a default, or whether you're stating that this setting is already enabled.

Answer to the former:
I personal feel that this would be a very useful default setting.

Answer to the latter:
I certainly found that it was *not* enabled by default. EG a record with the word "Māori": Searches for "Maori" would return zero results, searches for "Māori" would return many results.

dhensby:
I'm asking for thoughts on whether we should add this setting so it's on by default.

> [edit: but, enabling it now could cause problems for others]

What problems do you envisage?

chrispenny:
I guess when I say "problems", I just mean that it is possible that some projects using this module might be purposefully excluding macrons (or vice versa) from their search results.

dhensby:
I shouldn't block this merge based on whether it should or shouldn't be default... I'd be surprised if people were deliberately blocking UTF-8 chars, especially more than the people who want them need to add this...

PR #143, opened 2017-06-15
Index now supports multiple relations with the same name

mfendeksilverstripe:
Example setup:

`Page1` - many many - `Tags`
`Page2` - many many - `Tags`

Let's assume both Pages extend Page. Both pages have a relation `Tags`, but for same reason this relation has to be defined in each `Page1` and `Page2` and can't be moved to a common parent. We are interested in building index on `Tags.Title` and class added to index is `Page`.

Old behaviour when building index:

Page types are listed and first one that has relation `Tags` is picked. The rest of pages types are ignored and will not be present in the search index.

New behaviour (after this change):

All Page types that have relation `Tags` defined in them are listed and added to index. Pages that have the `Tags` relation by inheritance only will not be listed in the index.

### DIFF
--- a/code/search/SearchIndex.php
+++ b/code/search/SearchIndex.php
@@ -37,6 +37,27 @@ abstract class SearchIndex extends ViewableData
      */
     private static $hide_ancestor;
 
+    /**
+     * Used to separate class name and relation name in the sources array
+     * this string must not be present in class name
+     * @var string
+     * @config
+     */
+    private static $class_delimiter = '_|_';
+
+    /**
+     * This is used to clean the source name from suffix
+     * suffixes are needed to support multiple relations with the same name on different page types
+     * @param string $source
+     * @return string
+     */
+    private function getSourceName($source)
+    {
+        $source = explode(self::$class_delimiter, $source);
+
+        return $source[0];
+    }
+
     public function __construct()
     {
         parent::__construct();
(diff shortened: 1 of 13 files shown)


code/search/SearchIndex.php
```diff
@@ -77,26 +98,49 @@ public function fieldData($field, $forceType = null, $extraOptions = array())
...
-                    foreach (SearchIntrospection::hierarchy($source, $options['include_children']) as $dataclass) {
+                    foreach (SearchIntrospection::hierarchy($source, $baseOptions['include_children']) as $dataclass) {
+                        $class = null;
+                        $options = $baseOptions;
                         $singleton = singleton($dataclass);
 
                         if ($hasOne = $singleton->has_one($lookup)) {
+                            // we only want to include base class for relation, omit classes that inherited the relation
+                            $relationList = Config::inst()->get($dataclass, 'has_one', Config::UNINHERITED);
+                            $relationList = (!is_null($relationList)) ? $relationList : [];
```
robbieaverill: We'll need to support PHP 5.3, so will need old school array declaration

code/search/SearchIndex.php
```diff
@@ -109,8 +153,10 @@ public function fieldData($field, $forceType = null, $extraOptions = array())
                             if (!isset($options['origin'])) {
                                 $options['origin'] = $dataclass;
                             }
-                            $next[$class] = $options;
-                            continue 2;
+
+                            // we add suffix here to prevent the relation to be overwritten by other instances
+                            // all sources lookups must clean the source name before reading it via getSourceName()
+                            $next[$class . self::$class_delimiter . $dataclass] = $options;
```
robbieaverill: You've declared `$class_delimiter` as a configurable property, so might be worth using `self::config()->get('class_delimiter')` here instead of accessing the property directly

PR #144, opened 2017-06-15
Index now supports multiple relations with the same name

mfendeksilverstripe:
Example setup:

`Page1` - many many - `Tags`
`Page2` - many many - `Tags`

Let's assume both Pages extend `Page`. Both pages have a relation `Tags`, but for same reason this relation has to be defined in each `Page1` and `Page2` and can't be moved to a common parent. We are interested in building index on `Tags.Title` and class added to index is `Page`.

Old behaviour when building index:

Page types are listed and first one that has relation `Tags` is picked. The rest of pages types are ignored and will not be present in the search index.

New behaviour (after this change):

All Page types that have relation `Tags` defined in them are listed and added to index. Pages that have the `Tags` relation by inheritance only will not be listed in the index.

### DIFF
--- a/code/search/SearchUpdater.php
+++ b/code/search/SearchUpdater.php
@@ -97,6 +97,7 @@ class SearchUpdater extends Object
             $id = $details['id'];
             $state = $details['state'];
             $class = $details['class'];
+            $command = $details['command'];
             $fields = isset($details['fields']) ? $details['fields'] : array();
 
             $base = ClassInfo::baseDataClass($class);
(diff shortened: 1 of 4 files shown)


robbieaverill:
@mfendeksilverstripe I put a couple of comments on #143 FYI

mfendeksilverstripe:
Thanks, it's fixed now.

dhensby:
This is a new feature/API so it should go against the 2 branch

dhensby:
gah - I gone goofed up - I'll open a new PR

dhensby:
@mfendeksilverstripe FYI - you should create new branches when you make a PR rather than going off of your own 2.3 / master / etc branches because this give you flexibility to rebase/amend at will.

see: https://docs.silverstripe.org/en/4/contributing/code/#step-by-step-from-forking-to-sending-the-pull-request for some guidance

PR #145, opened 2017-06-16
Index now supports multiple relations with the same name.

dhensby:
Replaces #144

### DIFF
--- a/code/search/SearchIndex.php
+++ b/code/search/SearchIndex.php
@@ -37,6 +37,27 @@ abstract class SearchIndex extends ViewableData
      */
     private static $hide_ancestor;
 
+    /**
+     * Used to separate class name and relation name in the sources array
+     * this string must not be present in class name
+     * @var string
+     * @config
+     */
+    private static $class_delimiter = '_|_';
+
+    /**
+     * This is used to clean the source name from suffix
+     * suffixes are needed to support multiple relations with the same name on different page types
+     * @param string $source
+     * @return string
+     */
+    private function getSourceName($source)
+    {
+        $source = explode(self::config()->get('class_delimiter'), $source);
+
+        return $source[0];
+    }
+
     public function __construct()
     {
         parent::__construct();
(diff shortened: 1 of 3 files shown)


code/search/SearchIndex.php
```diff
@@ -37,6 +37,27 @@
...
+     */
+    private static $class_delimiter = '_|_';
+
+    /**
+     * This is used to clean the source name from suffix
+     * suffixes are needed to support multiple relations with the same name on different page types
+     * @param string $source
+     * @return string
+     */
+    private function getSourceName($source)
```
chillu: We don't make functions private usually, please mark it protected unless there's a really good reason that a subclass shouldn't be able to alter this behaviour

code/search/SearchIndex.php
```diff
@@ -77,26 +98,49 @@ public function fieldData($field, $forceType = null, $extraOptions = array())
...
+                    $source = $this->getSourceName($source);
 
-                    foreach (SearchIntrospection::hierarchy($source, $options['include_children']) as $dataclass) {
+                    foreach (SearchIntrospection::hierarchy($source, $baseOptions['include_children']) as $dataclass) {
+                        $class = null;
+                        $options = $baseOptions;
                         $singleton = singleton($dataclass);
 
                         if ($hasOne = $singleton->has_one($lookup)) {
+                            // we only want to include base class for relation, omit classes that inherited the relation
```
chillu: Who do we exclude inherited relations here? That seems like a recipe for a lot of dev confusion. From my understanding, we do include inherited fields, e.g. if you use `$myIndex->addClass('Page')` and `$myIndex->addField('Content')`, it'll index `Content` even though it's defined on the parent class (`SiteTree`)

code/search/SearchIndex.php
```diff
@@ -77,26 +98,49 @@ public function fieldData($field, $forceType = null, $extraOptions = array())
             foreach ($lookups as $lookup) {
                 $next = array();
 
-                foreach ($sources as $source => $options) {
-                    $class = null;
+                foreach ($sources as $source => $baseOptions) {
```
chillu: Please update the `fieldData()` method docs to note on how it handles relationships

code/search/SearchIndex.php
mfendeksilverstripe: I don't think there is a functional difference between adding the inherited relations or not as the index will fall back on the parent class relation. The difference would be the content of the XML which is output of this process. We probably don't want to have the XML full of fields that have no impact on the search.

Of course I'm happy to discuss this further as I'm quite new to this module.

mfendeksilverstripe:
@dhensby Can you please give me write access to this fork? I would like to add some changes based on feedback from Ingo.

mfendeksilverstripe:
@dhensby @chillu I processed the feedback and made some new changes. I also includes Ingo's tests and my own as well. I pushed the new version to my own fork because I don't have access to this one. I also addressed one of the issues Ingo raised (include classes that inherit the relation). Such classes do not need to be included in the index because their parent class will be included instead (regardless if parent class was originally meant to be included in the index or not). This is because the inner loop calls `hierarchy` which will include the parent class if needed.

https://github.com/mfendeksilverstripe/silverstripe-fulltextsearch/commit/32a6cfa8030ef7c091f1880fe1a7ff510f8923ab

Please have a look and let me know if you have more feedback for me.

dhensby:
I've integrated these changes.

If you need to make further changes either @chillu can push to my branch (he has permissions) or feel free to close this PR and open a new one. I'm not sure how to add permissions to individual branches to specific users.

PR #148, opened 2017-08-18
Set SearchUpdateCommitJobProcessor::$dirty_indexes default to array, not bool. Fixes #151

### LINKED ISSUE #151: v2.3.1: $dirty_indexes incorrect default property value
Fix:
https://github.com/silverstripe/silverstripe-fulltextsearch/pull/148

Currently in `SearchUpdateCommitJobProcessor`:
```php
/**
 * List of dirty indexes to be committed
 *
 * @var array
 */
public static $dirty_indexes = true;
```

This results in many Warning being thrown as `$dirty_indexes` is used (from what I could see) exclusively as type `array`, not `boolean`.

EG:
```
#0 in_array() expects parameter 2 to be array, boolean given:: called at [/sites/statsunleashed/releases/20170811014625/fulltextsearch/code/search/processors/SearchUpdateCommitJobProcessor.php:217]
```

chrispenny:
Related issue:
https://github.com/silverstripe/silverstripe-fulltextsearch/issues/151

Currently in `SearchUpdateCommitJobProcessor`:
```php
/**
 * List of dirty indexes to be committed
 *
 * @var array
 */
public static $dirty_indexes = true;
```

This results in many Warning being thrown as `$dirty_indexes` is used (from what I could see) exclusively as type `array`, not `boolean`.

EG:
```
#0 in_array() expects parameter 2 to be array, boolean given:: called at [/sites/statsunleashed/releases/20170811014625/fulltextsearch/code/search/processors/SearchUpdateCommitJobProcessor.php:217]
```

### DIFF
--- a/code/search/processors/SearchUpdateCommitJobProcessor.php
+++ b/code/search/processors/SearchUpdateCommitJobProcessor.php
@@ -8,7 +8,7 @@ class SearchUpdateCommitJobProcessor implements QueuedJob
 {
     /**
      * The QueuedJob queue to use when processing commits
-     * 
+     *
      * @config
      * @var int
      */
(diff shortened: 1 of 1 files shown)


dhensby:
happy @robbieaverill ?

robbieaverill:
Yep

PR #150, opened 2017-09-18
Add nested filtering via Criteria and Criterion objects.

chrispenny:
# Purpose

Currently in `fulltextsearch` we can `filter` and we can `exclude`, but both of these are "single level", in that you cannot group filters with combinations of `AND` and `OR` statements.

In order to allow for more complex filtering, I have implemented criterion objects. These objects can be joined and grouped in any number of ways - allowing you to achieve any kind of nested filter that you might want.

# New objects

* **`SearchCriteriaInterface`**: Interface for `SearchCriterion` and `SearchCriteria` classes.
* **`SearchCriterion`**: An object containing a single field filter (target field, comparison value, comparison type).
* **`SearchCriteria`**: An object containing a collection of `SearchCriterion` and/or `SearchCriteria` with conjunctions (IE: `AND`, `OR`) between each.
* **`SearchQueryWriter`**: A class used to generate a query string based on a `SearchCriterion`.
* **`SearchAdapterInterface`**: An Interface for our Search. This adapter will control what `SearchQueryWriter` is used for each `SearchCriteria`.

# General usage

We need 3 things to create a `SearchCriterion`:

* **`Target`**: EG the field in our Search Index that we want to filter against.
* **`Value`**: The value we want to use for comparison.
* **`Comparison`**: The type of comparison (EG: `EQUAL`, `IN`, etc).

All currently supported comparisons can be found as constants in `SearchCriterion`.

# Creating a new `SearchCriterion`

## Method 1a and 1b
```php
// `EQUAL` is the default comparison for `SearchCriterion`, so no third param is required.
$criterion = new SearchCriterion('Product_Title', 'My Product');

// Or use the `create` static method.
$criterion = SearchCriterion::create('Product_Title', 'My Product');
```

# Creating a new `SearchCriteria`

`SearchCriteria` has a property called `$clauses` which is a collection of `SearchCriterion` (above) and/or `SearchCriteria` (allowing for infinite nesting of clauses), along with the conjunction used between each clause (IE: `AND`, `OR`). We want to build up our `SearchCriteria` by adding to it's `$clauses` collection.

`SearchCriteria` can either be passed an object that implements `SearchCriteriaInterface`, or it can be passed the `Target`, `Value`, and `Comparison` (like above).

## Method 1
Instantiate a new `SearchCriteria` by providing an already instantiated `SearchCriterion` object. This `$criterion` will be added as the first item in the `$clauses` collection.
```php
$criteria = SearchCriteria::create($criterion);
```

## Method 2
Instantiate a new `SearchCriteria` objects and define the `Target`, `Value`, and `Comparison`. `SearchCriteria` will create a new `SearchCriterion` object based on the values, and add it to the `$clauses` collection.
```php
$criteria = SearchCriteria::create('Product_CatID', array(21, 24, 25), SearchCriterion::IN);
```

# Adding additional `SearchCriterion` to our `SearchCriteria`
When you want to add more complexity to your `SearchCriteria`, there are two methods available:

* `addAnd`: Add a new `SearchCriterion` or `SearchCriteria` with an `AND` conjunction.
* `addOr`: Add a new `SearchCriterion` or `SearchCriteria` with an `OR` conjunction.

## Method 1
Use method chaining to create a `SearchCriterion` with two clauses.
```php
// Filter by products with stock that are in either of these 3 categories.
$criteria = SearchCriteria::create('Product_CatID', array(21, 24, 25), SearchCriterion::IN)
    ->addAnd('Product_Stock', 0, SearchCriterion::GREATER_THAN);
```

## Method 2
Systematically add clauses to your already instantiated `SearchCriteria`.
```php
// Filter by products in either of these 3 categories.
$criteria = SearchCriteria::create('Product_CatID', array(21, 24, 25), SearchCriterion::IN);

... other stuff

// Filter by products with stock.
$criteria->addAnd('Product_StockLevel', 0, SearchCriterion::GREATER_THAN);
```

# Adding multiple levels of filtering to our `SearchCriteria`

`SearchCriteria` also allows you to pass in other `SearchCriteria` objects as you instantiate it and as you use the `addAnd` and `addOr` methods.
```php
// Filter by products that are in either of these 3 categories with stock.
$stockCategoryCriteria = SearchCriteria::create('Product_CatID', array(21, 24, 25), SearchCriterion::IN)
    ->addAnd('Product_Stock', 0, SearchCriterion::GREATER_THAN);

// Filter by products in Category ID  1 with stock over 5.
$legoCriteria = SearchCriteria::create('Product_CatID', 1, SearchCriterion::EQUAL)
    ->addAnd('Product_Stock', 5, SearchCriterion::GREATER_THAN);

// Combine the two criteria with an `OR` conjunction
$criteria = SearchCriteria::create($stockCategoryCriteria)
    ->addOr($legoCriteria);
```
The end result is that you get any products that are in categories 21, 24, 25 that have stock, and any products in category 1 with more than 5 in stock.

# Adding `SearchCriteria` to our `SearchQuery`
Our `SearchQuery` class now has a property called `$criteria` which holds all of our `SearchCriteria`. You can add new `SearchCriteria` by using `SearchQuery::filterBy()`.

## Method 1
Pass in an already instantiated `SearchCriteria` object. If you implemented complex filtering (above), you will probably need to follow this method - fully creating your `SearchCriteria` first, and then passing it to the `SearchQuery`.
```
$query->filterBy($criteria);
```

## Method 2a
Where basic (single level) filtering is ok, the `SearchQuery::filterBy()` method can be used to create your `SearchCriterion` and `SearchCriteria` object.
```
$query->filterBy('Product_CatID', array(21, 24, 25), SearchCriterion::IN);
```

## Method 2b
The `filterBy()` method will return the **current** `SearchCriteria`, this allows you to method chain the `addAnd` and `addOr` methods.
```php
// Filter by products with stock that are in either of these 3 categories.
$searchQuery->filterBy('Product_CategoryID', array(21, 24, 25), SearchCriterion::IN)
            ->addAnd('Product_StockLevel', 0, SearchCriterion::GREATER_THAN);
```

Each item in the `$criteria` collection are treated with an `AND` conjunction (matching current `filter`/`exclude` functionality).

# Search Query Writers

Provided are 3 different `SearchQueryWriter`s for Solr:

* `SolrSearchQueryWriter_Basic`
* `SolrSearchQueryWriter_In`
* `SolrSearchQueryWriter_Range`

When these Writers are provided a `SearchCriterion`, they will generate the desired query string.

# Search Adapters

Search Adapters need to provide the following information:

* What is the search engine's conjunction strings? (EG: are they "AND" and "OR", or are they "&&" and "||", etc).
* What is the desired comparison container string? (EG: "**+(** query here **)**") for Solr).
* Most importantly - how to generate the query string from a `SearchCriterion`.

The `SolrSearchAdapter` uses `SearchQueryWriter`s (above) to generate query strings from a `SearchCriterion`.

I realise that having `SearchQueryWriter`s is an additional step, and that usually the `Adapter` would do the generating (all by itself), but I really wanted to give devs a way to bypass the default "writing" behaviour if they needed to.

# Customising your `SearchCriterion`/`SearchQueryWriter`

If you find that you do not want your `SearchCriterion` being parsed by one of the default `SearchQueryWriter`s (for whatever reason), you can optionally pass your own `SearchQueryWriter` to your `SearchCriterion` either as the **fourth parameter** when instantiating it, or by calling `setSearchQueryWriter()`.

If this value is set, then the (default Solr) Adapter will always use the provided `SearchQueryWriter`, rather than deciding for itself.

This should allow you to have full control over how your query strings are being generated if the default `SearchQueryWriter`s are not cutting it for you.

### DIFF
--- a/_config/config.yml
+++ b/_config/config.yml
@@ -4,3 +4,8 @@ Name: fulltextsearchconfig
 SilverStripe\ORM\DataObject:
   extensions:
     - SilverStripe\FullTextSearch\Search\Extensions\SearchUpdater_ObjectHandler
+
+SilverStripe\Core\Injector\Injector:
+  SilverStripe\FullTextSearch\Search\Queries\SearchQuery:
+    calls:
+      - [ setHandler, [ %$SilverStripe\FullTextSearch\Search\Adapters\SolrSearchAdapter ]]
--- a/docs/en/00_index.md
+++ b/docs/en/00_index.md
@@ -19,6 +19,7 @@
     - [Searching value ranges](04_querying.md#searching-value-ranges)
     - [Empty or existing values](04_querying.md#empty-or-existing-values)
     - [Executing your query](04_querying.md#executing-your-query)
+    - [Complex filtering with Criteria](04_querying.md#complex-filtering-with-criteria)
 - Advanced configuration
     - [Facets](05_advanced_configuration.md#facets)
     - [Using multiple indexes](05_advanced_configuration.md#multiple-indexes)
--- a/docs/en/04_querying.md
+++ b/docs/en/04_querying.md
@@ -128,3 +128,164 @@ The return value of a `search()` call is an object which contains a few properti
  * `Suggestion`: (optional) Any suggested spelling corrections in the original query notation
  * `SuggestionNice`: (optional) Any suggested spelling corrections for display (without query notation)
  * `SuggestionQueryString` (optional) Link to repeat the search with suggested spelling corrections
+
+## Complex Filtering with Criteria
+ 
+### Filtering related Objects
+ 
+* `SearchCriteriaInterface`: Interface for `SearchCriterion` and `SearchCriteria` classes.
+* `SearchCriterion`: An object containing a single field filter (target field, comparison value, comparison type).
+* `SearchCriteria`: An object containing a collection of `SearchCriterion` and/or `SearchCriteria` with conjunctions (IE: `AND`, `OR`) between each.
+* `SearchQueryWriter`: A class used to generate a query string based on a `SearchCriterion`.
+* `SearchAdapterInterface`: An Interface for our SearchAdapters. This adapter will control what `SearchQueryWriter` is used for each `SearchCriteria`.
+ 
+### General usage
+ 
+We need 3 things to create a `SearchCriterion`:
+ 
+* **`Target`**: EG the field in our Search Index that we want to filter against.
+* **`Value`**: The value we want to use for comparison.
+* **`Comparison`**: The type of comparison (EG: `EQUAL`, `IN`, etc).
+ 
+All currently supported comparisons can be found as constants in `SearchCriterion`.
+ 
+### Creating a new `SearchCriterion`
+ 
+#### Method 1a and 1b
+ 
+```php
+// `EQUAL` is the default comparison for `SearchCriterion`, so no third param is required.
+$criterion = new SearchCriterion('Product_Title', 'My Product');
+ 
+// Or use the `create` static method.
+$criterion = SearchCriterion::create('Product_Title', 'My Product');
+```
+ 
+### Creating a new `SearchCriteria`
+ 
+`SearchCriteria` has a property called `$clauses` which is a collection of `SearchCriterion` (above) and/or `SearchCriteria` (allowing for infinite nesting of clauses), along with the conjunction used between each clause (IE: `AND`, `OR`). We want to build up our `SearchCriteria` by adding to it's `$clauses` collection.
+ 
+`SearchCriteria` can either be passed an object that implements `SearchCriteriaInterface`, or it can be passed the `Target`, `Value`, and `Comparison` (like above).
+ 
+#### Method 1
+ 
+Instantiate a new `SearchCriteria` by providing an already instantiated `SearchCriterion` object. This `$criterion` will be added as the first item in the `$clauses` collection.
+ 
+```php
+$criteria = SearchCriteria::create($criterion);
+```
+ 
+#### Method 2
+ 
+Instantiate a new `SearchCriteria` objects and define the `Target`, `Value`, and `Comparison`. `SearchCriteria` will create a new `SearchCriterion` object based on the values, and add it to the `$clauses` collection.
+ 
+```php
+$criteria = SearchCriteria::create('Product_CatID', array(21, 24, 25), AbstractCriterion::IN);
+```
+ 
+### Adding additional `SearchCriterion` to our `SearchCriteria`
+ 
+When you want to add more complexity to your `SearchCriteria`, there are two methods available:
+ 
+* `addAnd`: Add a new `SearchCriterion` or `SearchCriteria` with an `AND` conjunction.
+* `addOr`: Add a new `SearchCriterion` or `SearchCriteria` with an `OR` conjunction.
+ 
+#### Method 1
+ 
+Use method chaining to create a `SearchCriterion` with two clauses.
+ 
+```php
+// Filter by products with stock that are in either of these 3 categories.
+$criteria = SearchCriteria::create('Product_CatID', array(21, 24, 25), AbstractCriterion::IN)
+    ->addAnd('Product_Stock', 0, AbstractSearchCriterion::GREATER_THAN);
+```
+ 
+#### Method 2
+ 
+Systematically add clauses to your already instantiated `SearchCriteria`.
+ 
+```php
+// Filter by products in either of these 3 categories.
+$criteria = SearchCriteria::create('Product_CatID', array(21, 24, 25), AbstractCriterion::IN);
+ 
+... other stuff
+ 
+// Filter by products with stock.
+$criteria->addAnd('Product_StockLevel', 0, AbstractCriterion::GREATER_THAN);
+```
+ 
+### Adding multiple levels of filtering to our `SearchCriteria`
+ 
+`SearchCriteria` also allows you to pass in other `SearchCriteria` objects as you instantiate it and as you use the `addAnd` and `addOr` methods.
+ 
+```php
+// Filter by products that are in either of these 3 categories with stock.
+$stockCategoryCriteria = SearchCriteria::create('Product_CatID', array(21, 24, 25), AbstractCriterion::IN)
+    ->addAnd('Product_Stock', 0, AbstractSearchCriterion::GREATER_THAN);
+ 
+// Filter by products in Category ID  1 with stock over 5.
+$legoCriteria = SearchCriteria::create('Product_CatID', 1, AbstractCriterion::EQUAL)
+    ->addAnd('Product_Stock', 5, AbstractSearchCriterion::GREATER_THAN);
+ 
+// Combine the two criteria with an `OR` conjunction
+$criteria = SearchCriteria::create($stockCategoryCriteria)
+    ->addOr($legoCriteria);
+```
+ 
+### Adding `SearchCriteria` to our `SearchQuery`
+ 
+Our `SearchQuery` class now has a property called `$criteria` which holds all of our `SearchCriteria`. You can add new `SearchCriteria` by using `SearchQuery::filterBy()`.
+ 
+#### Method 1
+ 
+Pass in an already instantiated `SearchCriteria` object. If you implemented complex filtering (above), you will probably need to follow this method - fully creating your `SearchCriteria` first, and then passing it to the `SearchQuery`.
+ 
+```php
+$query->filterBy($criteria);
+```
+ 
+#### Method 2a
+Where basic (single level) filtering is ok, the `SearchQuery::filterBy()` method can be used to create your `SearchCriterion` and `SearchCriteria` object.
+ 
+```php
+$query->filterBy('Product_CatID', array(21, 24, 25), AbstractCriterion::IN);
+```
+ 
+#### Method 2b
+The `filterBy()` method will return the **current** `SearchCriteria`, this allows you to method chain the `addAnd` and `addOr` methods.
+ 
+```php
+// Filter by products with stock that are in either of these 3 categories.
+$searchQuery->filterBy('Product_CategoryID', array(21, 24, 25), AbstractCriterion::IN)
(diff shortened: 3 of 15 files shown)


madmatt:
I can't +1 this enough. I think this is awesome, but it's obviously not BC, so we'd need to release a new major version for it.

Thoughts on doing this @robbieaverill?

chrispenny:
@madmatt I guess this would need to be updated for v2.4, but why is this not BC? I haven't changed any existing functionality - only added new methods that folks can start to use if they choose. I don't think this would break any existing v2.3 installations.

silverstripe-elliot:
It's backwards compatible because using SearchCriterion is optional. Existing code should work fine

madmatt:
Ah sorry, I had a quick squiz and thought that you had replaced the existing query interface with the new criterions - my fault for not reading the PR fully!

andrewandante:
Pretty keen to get this merged in so we can push it up through to the 3.x line, and start looking at deprecating the existing `addFilter` and `addExclude` methods

code/search/SearchCriteria.php
```diff
@@ -0,0 +1,233 @@
...
+     * @param string
+     */
+    const UND = 'AND';
+
+    /**
+     * "OR" is a protected PHP7 keyword.
+     *
+     * @param string
+     */
+    const ODER = 'OR';
```
robbieaverill: I dislike the naming of these constants. Can you use `CONJUNCTION_OR` and `CONJUNCTION_AND` instead maybe?

code/search/SearchCriteria.php
```diff
@@ -0,0 +1,233 @@
...
+    public function setAdapter(SearchAdapterInterface $adapter)
+    {
+        $this->adapter = $adapter;
+
+        return $this;
+    }
+
+    /**
+     * @param string $ps Current prepared statement.
+     * @return void;
```
robbieaverill: The semi-colon isn't necessary here

code/search/SearchCriteria.php
```diff
@@ -0,0 +1,233 @@
...
+     * @param int $key
+     * @return string|null
+     */
+    protected function getConjunction($key)
+    {
+        if (!array_key_exists($key, $this->getConjunctions())) {
+            return null;
+        }
+
+        return $this->getConjunctions()[$key];
```
robbieaverill: This module still supports PHP 5.3, which doesn't support array dereferencing. Could you please assign `$this->getConjunctions()` to a variable and use that?

code/search/SearchCriteriaInterface.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    /**
+     * The method used in all SearchCriterion to generate and append their filter query statements.
+     *
+     * This is also used in SearchCriteria to loop through it's collected SearchCriterion and append the above. This
+     * allows us to have SearchCriteria and SearchCriterion in the same collections (allowing us to have complex nested
+     * filtering).
+     *
+     * @param $ps
+     * @return mixed
```
robbieaverill: Can we be stricter about this?

code/search/SearchCriterion.php
```diff
@@ -0,0 +1,274 @@
...
+    public static function create()
+    {
+        $args = func_get_args();
+
+        $class = get_called_class();
+        if ($class == 'Object') {
+            $class = array_shift($args);
+        }
+
+        return Injector::inst()->createWithArgs($class, $args);
```
robbieaverill: Is there a reason we can't just use `::create()` directly from Object/injector for this over defining our own `create` methods?

code/search/SearchCriterion.php
```diff
@@ -0,0 +1,274 @@
...
+    /**
+     * @return string
+     */
+    public function getComparison()
+    {
+        return $this->comparison;
+    }
+
+    /**
+     * @param $comparison
```
robbieaverill: Missing type

code/search/SearchCriterion.php
```diff
@@ -0,0 +1,274 @@
...
+    /**
+     * @return string
+     */
+    public function getTarget()
+    {
+        return $this->target;
+    }
+
+    /**
+     * @param $target
```
robbieaverill: Missing type

code/search/SearchCriterion.php
```diff
@@ -0,0 +1,274 @@
...
+    /**
+     * @return mixed
+     */
+    public function getValue()
+    {
+        return $this->value;
+    }
+
+    /**
+     * @param $value
```
robbieaverill: Missing type (`mixed` as per getter and class prop?)

code/search/SearchQuery.php
```diff
@@ -21,19 +21,33 @@ class SearchQuery extends ViewableData
     public $require = array();
     public $exclude = array();
 
+    /**
+     * @var SearchCriteriaInterface[]
+     */
+    public $criteria = array();
```
robbieaverill: I think using public properties here is something we'd like to move away from. Can we use getters and setters instead and make this protected?

code/solr/SolrSearchAdapter.php
```diff
@@ -0,0 +1,88 @@
...
+     * @return string
+     * @throws InvalidArgumentException
+     */
+    public function getConjunctionFor($conjunction)
+    {
+        switch ($conjunction) {
+            case SearchCriteria::UND:
+                return ' AND ';
+            case SearchCriteria::ODER:
+                return ' OR ';
```
robbieaverill: This is duplicating the constant values. How would you feel about this instead?
```php
case SearchCriteria::UND:
case SearchCriteria::ODER:
    return sprintf(' %s ', $conjunction);
```

Note: I commented earlier about the naming of these constants

code/solr/SolrSearchAdapter.php
```diff
@@ -0,0 +1,88 @@
...
+            return $searchCriterion->getSearchQueryWriter();
+        }
+
+        switch ($searchCriterion->getComparison()) {
+            case SearchCriterion::EQUAL:
+            case SearchCriterion::NOT_EQUAL:
+                return new SolrSearchQueryWriter_Basic();
+            case SearchCriterion::IN:
+            case SearchCriterion::NOT_IN:
+                return new SolrSearchQueryWriter_In();
```
robbieaverill: Can you please use injector for this instantiation, in case anyone wants to extend them?

code/solr/searchquerywriters/SolrSearchQueryWriter_Basic.php
```diff
@@ -0,0 +1,46 @@
+<?php
+
+class SolrSearchQueryWriter_Basic extends AbstractSearchQueryWriter
```
robbieaverill: If possible, please try and avoid using underscores in class names (for future proofing against PSR-2)

code/solr/searchquerywriters/SolrSearchQueryWriter_Basic.php
```diff
@@ -0,0 +1,46 @@
...
+     * @return string
+     */
+    public function generateQueryString(SearchCriterion $searchCriterion)
+    {
+        $qs = $this->getComparisonPolarity($searchCriterion->getComparison());
+        $qs .= '(';
+        $qs .= $searchCriterion->getTarget();
+        $qs .= $this->getComparisonConjunction();
+        $qs .= $searchCriterion->getQuoteValue($searchCriterion->getValue());
+        $qs .= ')';
```
robbieaverill: I feel like something like this might look a little neater:
```php
return sprintf(
    '%s(%s%s%s)',
    $this->getComparis...,
    $searchCrit...,
     $this->getCompar....
    $searchCriter.....
);
```

What do you think? It's probably personal preference 😆

code/solr/searchquerywriters/SolrSearchQueryWriter_Range.php
```diff
@@ -0,0 +1,141 @@
...
+    {
+        $qs = $this->getComparisonPolarity($searchCriterion->getComparison());
+        $qs .= '(';
+        $qs .= $searchCriterion->getTarget();
+        $qs .= $this->getOpenComparisonContainer($searchCriterion->getComparison());
+        $qs .= $this->getLeftComparison($searchCriterion);
+        $qs .= $this->getComparisonConjunction();
+        $qs .= $this->getRightComparison($searchCriterion);
+        $qs .= $this->getCloseComparisonContainer($searchCriterion->getComparison());
+        $qs .= ')';
```
robbieaverill: See previous comment about how this is assembled

docs/en/index.md
```diff
@@ -240,7 +240,165 @@ Example:
 	);
 	$results = singleton('MyIndex')->search($query);
 
-## Filtering
+## Basic Filtering
+
+Basic filtering is provided through the "filter" and "exclude" methods:
+
+	// Filter by products in either of these 3 categories.
```
robbieaverill: Please use GitHub flavoured markdown for code block examples (three backticks)

chrispenny:
Woooo! Thanks so much for the feedback @robbieaverill. I'm very excited to be able to get back into this PR.

I'm assuming we get a hackday next month? Most likely that will be the next chunk of time I have available, but I'll see what OS time might be available in our sprint/s.

chrispenny:
@robbieaverill does it make more sense for me to update this PR to merge into `v3.x`, or do we want this available in `v2.x` as well?

I had actually already started the process of `SS4'ing` my changes, so it would be minimal extra effort for me to just continue with that.

robbieaverill:
Up to you :)

chrispenny:
@robbieaverill Thanks again for all of the feedback. I've processed it all now, and I've upgraded the PR to SS4.

I've set the base to `dev-master`, but let me know if it's better for this to go against `3.3`.

I've added a bunch of tests that check that the correct query string is output by each of the `Writers` and `SolrIndex`. I'm not sure if there's much else I can test for that would be meaningful? Checking that the query string actually performed as expected would require a non-faked query (I assume?).

robbieaverill:
Re-running the Travis builds to ensure we're still green then will merge

chrispenny:
Woohoo! Thanks so much, @robbieaverill! I appreciate all the feedback/time you gave for this PR 😃

madmatt:
This is huuuuuge. Thanks for all the hard work @chrispenny and @robbieaverill!

Issue #154, opened 2017-10-25
Solr_Reindex task with Translatable content - Cannot index multiple languages' content


BrewCurious:
I have a website running SS 3.6.x with Translatable module installed. By default, the Solr_Reindex task will only index SiteTree content within the current locale of the user's session.

I've hacked in an extension hook at the beginning and end of the run() method within the FullTextSearch's Solr_Reindex task (/fulltextsearch/code/solr/Solr.php) in order to toggle the translatable enable_locale_filter() method for each iteration of the reindex task. That works, but I'd like to find a better way to approach the problem that doesn't involve hacking the FTS module.

Since the reindex task is broken up by the recordsPerRequest value in the config, I haven't been able to figure out a way to toggle the translatable locale filter for all iterations of the task. It works for the first task, but gets reset for all subsequent iterations.

Could extend hooks be added to the beginning and end of the tasks to toggle and reinstate translatable filters? Or is there another way to provide filter settings that won't be replaced between multiple reindex tasks?

robbieaverill:
>Could extend hooks be added to the beginning and end of the tasks to toggle and reinstate translatable filters? 

Yep, that's fully reasonable!

BrewCurious:
Created pull request #155 for review.

BrewCurious:
As soon as I submitted the pull request I realized that there likely isn't a scenario where the hooks in the Solr_Configure task would be used since they don't (to my knowledge) get run multiple times in the way that the reindexing task does.

That could get removed unless anyone else could think of a reason that they might be useful.

robbieaverill:
@BrewCurious do you think #155 (released) fixes this issue?

PR #155, opened 2017-10-25
Adding extension hooks before and after Solr tasks

BrewCurious:
Adding extension hooks before and after Solr_Configure and Solr_Reindex tasks to allow other modules (e.g. Translatable) to modify filters consistently across multiple task iterations.

Addressing issue #154.

### DIFF
--- a/code/solr/Solr.php
+++ b/code/solr/Solr.php
@@ -227,6 +227,8 @@ class Solr_Configure extends Solr_BuildTask
     {
         parent::run($request);
 
+        $this->extend('updateBeforeSolrConfigureTask', $request);
+
         // Find the IndexStore handler, which will handle uploading config files to Solr
         $store = $this->getSolrConfigStore();
         $indexes = Solr::get_indexes();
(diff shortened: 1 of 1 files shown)


code/solr/Solr.php
```diff
@@ -227,6 +227,8 @@ public function run($request)
     {
         parent::run($request);
 
+        $this->extend('updateBeforeSolrConfigureTask', $request);
```
dhensby: `parent::run()` runs before this... Should this go first?

BrewCurious: I put it after the parent::run() call as I wasn't sure where it was in the code that the translatable filters were getting reset in my particular scenario and wanted to make sure that I was able to toggle that filter just before the actual index action took place.

BrewCurious:
OK... Git/PR noob question here, but is there anything that I should be doing to my PR to address the PHP < 5.6 issues in the Travis config?

robbieaverill:
@BrewCurious not a noob question at all! Travis has dropped support for PHP 5.3 on their trusty distros, so we've been temporarily adding `dist: precise` to the `.travis.yml` configuration to re-enable it (which you could do if you wanted to).

We'll probably be removing unsupported PHP versions from Travis builds at some stage next year.

robbieaverill:
I've updated the Travis configuration in the 2 branch (what the master branch was when you created this pull request) and have rebased your PR. Thanks for contributing @BrewCurious!

PR #156, opened 2017-11-10
Move SynonymFilterFactory to bottom of analyzer to include synonyms in search results

### LINKED ISSUE #153: Solr Synonyms Don't Work
**Scenario**
I create three different pages. 
On each of them I have put one of: gdp, jel, knp (unique words so I know these aren't anywhere else on the site)
I run a `Solr_Reindex`
I search for "gdp" and just that one page is returned in the results. 
Searching any of the unique words returns the correct result. 

Now, I'd like to make these words synonyms. 
I go into the `Fulltext Search` tab in `Settings` of the CMS (which uses CWP's `SynonymsSiteConfig`)
This config modifies the synonyms.txt file. 
I enter `gdp, jel, knp`as my Solr Synonyms. I run a Solr_Configure. 

**Result**
Now when I search any of the unique words, I get no results. 
I do a Reindex just to see if that has an effect. Still no results. 

**Expected Result**
What I would expect, is that if I searched for any of those words that *all 3 pages* would be returned. 

Am I misunderstanding the concept of Solr Synonyms? 
Or do they not function correctly in the module? 
An issue with version of Solr used in fulltextsearch-local? 
Or something else?

Just thought I would get the convo started and see if anyone else is having this issue. 😄

PsukheDelos:
Fixes #153

### DIFF
--- a/conf/solr/3/templates/types.ss
+++ b/conf/solr/3/templates/types.ss
@@ -140,7 +140,6 @@
       </analyzer>
       <analyzer type="query">
         <tokenizer class="solr.WhitespaceTokenizerFactory"/>
-        <filter class="solr.SynonymFilterFactory" synonyms="synonyms.txt" ignoreCase="true" expand="true"/>
         <filter class="solr.StopFilterFactory"
                 ignoreCase="true"
                 words="stopwords.txt"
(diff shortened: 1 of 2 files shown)


tractorcow:
Seems fine to me.

PR #173, opened 2017-11-15
FIX Replace Object reference with Injector and use ModuleLoader to resolve schema location

robbieaverill:
Allows Solr_Configure task to run


dhensby:
So we are green apart from the codecov target for this diff.

Is it ready @robbieaverill ?

robbieaverill:
Yep it's ready!

PR #180, opened 2017-11-29
API Make SearchVariant::call(...) arguments variadic

### LINKED ISSUE #165: Make SearchVariant::call(...) arguments variadic
With PHP 5.6 as the minimum for SS4 we can use [variadic arguments](http://php.net/manual/en/functions.arguments.php#functions.variable-arg-list) for things like this so simplify the method signature.

See comment: https://github.com/silverstripe/silverstripe-fulltextsearch/pull/141#discussion_r151013932

raissanorth:
Resolves #165


dhensby:
Yes please

PR #182, opened 2017-12-01
API Remove deprecated code

### LINKED ISSUE #169: Remove code that has been deprecated in version 2.0
With the 3.0 release we can remove code that was deprecated in version 2.0.

See comment: https://github.com/silverstripe/silverstripe-fulltextsearch/pull/141#discussion_r151015164

raissanorth:
Resolves #169

### DIFF
--- a/src/Solr/Tasks/Solr_Reindex.php
+++ b/src/Solr/Tasks/Solr_Reindex.php
@@ -1,4 +1,5 @@
 <?php
+
 namespace SilverStripe\FullTextSearch\Solr\Tasks;
 
 use ReflectionClass;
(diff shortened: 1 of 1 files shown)


src/Solr/Tasks/Solr_Reindex.php
```diff
@@ -121,38 +121,4 @@ protected function doReindex($request)
         $handler->triggerReindex($this->getLogger(), $this->config()->recordsPerRequest, $taskName, $class);
     }
 
-    /**
-     * @deprecated since version 2.0.0
-     */
-    protected function runFrom($index, $class, $start, $variantstate)
```
robbieaverill: This code is still being referenced, you'll need to remove the reference to it as well. It looks safe to kill.

https://github.com/silverstripe/silverstripe-fulltextsearch/blob/7b72774890c37c9fa0c2499e26ec5154e31a2000/src/Solr/Tasks/Solr_Reindex.php#L93-L101

robbieaverill: Also line 22 that suggests you can use the `start` argument when running the task - let's kill that too

PR #189, opened 2017-12-05
FIX Double escape namespace separators in class names when excluding fields

robbieaverill:
Before:

```php
$query->exclude(SiteTree::class . '_ShowInSearch', 0);
```

Result:
```
936953 [qtp2081303229-16] ERROR org.apache.solr.core.SolrCore  – org.apache.solr.common.SolrException: undefined field SilverStripeCMSModelSiteTree_ShowInSearch
```

Fix double escapes namespace separators in PHP class names when using them in `->exclude()` keys.

### DIFF
--- a/src/Solr/SolrIndex.php
+++ b/src/Solr/SolrIndex.php
@@ -941,7 +941,10 @@ abstract class SolrIndex extends SearchIndex
     {
         $fq = array();
         foreach ($searchQuery->exclude as $field => $values) {
-            $excludeq = array();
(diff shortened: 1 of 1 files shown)


madmatt:
So should we still be doing this to exclude something, or has the syntax changed?
```php
$query->exclude(SiteTree::class . '_ShowInSearch', 0);
```

robbieaverill:
Yep @madmatt that's correct

PR #195, opened 2017-12-21
API More TestState into source folder from tests

robbieaverill:
This was put into the tests namespace unintentionally but should live in the main module code


robbieaverill:
> Great job! Perhaps we should consider changing the namespace to emphasise that this is a TestState?

Yeah, good idea. That'll help us to differentiate this class from other *State classes that are injector services, rather than SapphireTest state classes

robbieaverill:
Updated

PR #200, opened 2018-02-15
NEW Use symfony/process to execute immediate system commands for reindexing

### LINKED ISSUE #197: System sake commands will not execute on SSP
`SolrReindexImmediateHandler::processGroup` runs a system command to run a Solr index immediately, however on SSP the shimmed environment variables don't transfer through to this command so it results in a "SilverStripe Framework requires database configuration defined via .env" or similar message:

```php
        $frameworkPath = ModuleLoader::getModule('silverstripe/framework')->getPath();
        $scriptPath = sprintf("%s%scli-script.php", $frameworkPath, DIRECTORY_SEPARATOR);
        $scriptTask = "php {$scriptPath} dev/tasks/{$taskName}";

        $cmd = "{$scriptTask} index={$indexClassEscaped} class={$class} group={$group} groups={$groups} variantstate={$statevar}";
        $cmd .= " verbose=1 2>&1";
        $logger->info("Running '$cmd'");

        // Execute script via shell
        $res = $logger ? passthru($cmd) : `$cmd`;
```

We should remove this and replace it with loading and executing the required task class instead. Ideally we'd replace the `$taskName` argument passes into `->triggerReindex()` with the instance of the task instead, so the immediate handler can simply call `$task->run($request)` after setting the requires data into the request object.

cc @silverstripe/core-team for thoughts on making this change

cc @elliot-sawyer @silverstripe-elliot @jakedaleweb

robbieaverill:
Resolves #197

### DIFF
--- a/src/Solr/Reindex/Handlers/SolrReindexImmediateHandler.php
+++ b/src/Solr/Reindex/Handlers/SolrReindexImmediateHandler.php
@@ -8,6 +8,7 @@ use SilverStripe\Core\Manifest\ModuleLoader;
 use SilverStripe\FullTextSearch\Solr\Solr;
 use SilverStripe\FullTextSearch\Solr\SolrIndex;
 use SilverStripe\ORM\DB;
+use Symfony\Component\Process\Process;
 
 /**
  * Invokes an immediate reindex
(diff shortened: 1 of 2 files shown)


src/Solr/Reindex/Handlers/SolrReindexImmediateHandler.php
```diff
@@ -77,11 +78,15 @@ protected function processGroup(
...
 
         $cmd = "{$scriptTask} index={$indexClassEscaped} class={$class} group={$group} groups={$groups} variantstate={$statevar}";
-        $cmd .= " verbose=1 2>&1";
+        $cmd .= " verbose=1";
         $logger->info("Running '$cmd'");
 
         // Execute script via shell
-        $res = $logger ? passthru($cmd) : `$cmd`;
+        $process = new Process($cmd);
+        $process->inheritEnvironmentVariables();
```
NightJar: Well that was alarmingly easy 👍

Issue #33, opened 2013-10-11
Solr searching fails when not searching Versioned content


madmatt:
If you create a sub-class of `SolrIndex` to define indexes on `DataObject`s that don't use the `Versioned` extension, the index will be built and loaded into Solr, but no searches performed against that index will work.

For example:

``` php
// Story.php
class Story extends DataObject {
    private static $db = array(
        'Title' => 'Varchar(255)'
    );
}

// StorySolrIndex.php
class StorySolrIndex extends SolrIndex {
    public function init() {
        $this->addClass('Story');
        $this->addFulltextField('Title');
    }
}
```

Running `Solr_Configure` and `Solr_Reindex` will work fine (`Solr_Reindex` will say `class Story, total: 5 in state []`)

However, creating an instance of that SolrIndex and running a query against it will fail:

``` php
$index = new StorySolrIndex();
$query = new SearchQuery();
$query->search('Test', null, array(
    'Story_Title' => 2
));

$results = $index->search($query);
```

Nothing will be returned, but Solr will log an Exception with the message:
`SEVERE: org.apache.solr.common.SolrException: undefined field _versionedstage`

If you add something to the `StorySolrIndex` that is versioned (e.g. `SiteTree`), then the search will work as usual, but with extra stuff in the index.

``` php
// StorySolrIndex.php
class StorySolrIndex extends SolrIndex {
    public function init() {
        $this->addClass('Story');
        $this->addClass('SiteTree'); // This works around the bug
        $this->addFulltextField('Title');
    }
}
```

madmatt:
@phptek, @kinglozzer, @unclecheese: A better workaround is to call ->inClass() on the SearchQuery object you pass through. In my original example, that would be:
```php
$query = new SearchQuery();
$query->inClass('Story');
```

So long as you only pass one class in, and that class doesn't have the `Versioned` extension on it, then it works fine.

[The offending line is here](https://github.com/silverstripe-labs/silverstripe-fulltextsearch/blob/ff62c2b9302d5fb94790969e5b2fd9a81b343b93/code/solr/SolrIndex.php#L368):
```php
SearchVariant::with(count($query->classes) == 1 ? $query->classes[0]['class'] : null)->call('alterQuery', $query, $this);
```

[...]

PR #202, opened 2018-02-23
NEW Add SearchVariant::withCommon to run callbacks on relevant variants rather than all

robbieaverill:
This pull request is an option for how we could address #33 - it adds a new method `SearchVariant::withCommon` which will return an array of search variants that are relevant to at least one of the classes given. SolrIndex can use this to provide either the classes in the index, or the classes specified for the query by user code.

This is different from `SearchVariant::with` which will apply all variants that are relevant to the environment if you don't provide it a specific class name (which `SolrIndex::search` does with a standard query). I believe this is what is causing #33, at least for me.

The unit test fix is actually an example of the bug, since `SearchUpdaterTest_Container` is not versioned so the versioned state is not relevant to it.

### DIFF
--- a/src/Search/Variants/SearchVariant.php
+++ b/src/Search/Variants/SearchVariant.php
@@ -2,11 +2,10 @@
 
 namespace SilverStripe\FullTextSearch\Search\Variants;
 
-use SilverStripe\ORM\DataObject;
+use ReflectionClass;
 use SilverStripe\Core\ClassInfo;
 use SilverStripe\Core\Config\Configurable;
 use SilverStripe\FullTextSearch\Utils\CombinationsArrayIterator;
-use ReflectionClass;
 
 /**
  * A Search Variant handles decorators and other situations where the items to reindex or search through are modified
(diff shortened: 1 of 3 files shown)


robbieaverill:
For what it’s with this change is non breaking so could be backported to SS3 branches

src/Search/Variants/SearchVariant.php
```diff
@@ -166,6 +166,42 @@ public static function with($class = null, $includeSubclasses = true)
...
+     * Similar to {@link SearchVariant::with}, except will only use variants that apply to at least one of the classes
+     * in the input array, where {@link SearchVariant::with} will run the query on the specific class you give it.
+     *
+     * @param array $classes
+     * @return SearchVariant_Caller
+     */
+    public static function withCommon(array $classes = [])
+    {
+        // Allow caching
+        $cacheKey = serialize($classes);
```
dhensby: is this an array of strings? if so can we change the docblock to `string[]` please.

I'd also recommend `md5` the `cacheKey` as a string representing a serialized array seems excessive for an array key.

robbieaverill: Yep, fair call. I'll update

NightJar:
whoop!

PR #204, opened 2018-02-23
FIX Don't assume a DataObject's table.

### LINKED ISSUE #168: Investigate todo comment re: failure if you addClass a non-table class
See comments https://github.com/silverstripe/silverstripe-fulltextsearch/pull/141#discussion_r151015408 and https://github.com/silverstripe/silverstripe-fulltextsearch/pull/141#discussion_r151015589

NightJar:
Knowing if a DataObject has a table or not is irrelevant - data is queried
out and pushed into solr all the same, whether there is a join table, a
base table, or any combination with maybe a few no-tables in between.
SomeDataClass::get() will fetch all data irrelevant of tables, that's the
point of an abstraction like SilverStripe's ORM.

The index featured is a default install, where `Page` is defined as

```php
use SilverStripe\CMS\Model\SiteTree;

class Page extends SiteTree
{

}
```

and defines an index [adding only `Page::class` to it](https://github.com/dnadesign/silverstripe-elemental/blob/master/src/Search/ElementalSolrIndex.php#L20).

capture

Resolves #168

### DIFF
--- a/src/Search/Indexes/SearchIndex.php
+++ b/src/Search/Indexes/SearchIndex.php
@@ -282,10 +282,6 @@ abstract class SearchIndex extends ViewableData
             throw new Exception('Can\'t add class to Index after fields have already been added');
         }
 
-        if (!DataObject::getSchema()->classHasTable($class)) {
-            throw new \InvalidArgumentException('Can\'t add classes which don\'t have data tables (no $db or $has_one set on the class)');
-        }
-
         $options = array_merge(array(
             'include_children' => true
         ), $options);


NightJar:
Fixes #168 
(once it has test)

NightJar:
Seems fair :)

dhensby:
wow - what is this about? what a bizarre constraint. I don't see that this could have any regression for existing code at all - this simply allows us to add all DO classes instead of an arbitrary subset of them.

PR #206, opened 2018-03-07
Fix issue where schema.xml etc is put in a different dir than the cor…

jakedaleweb:
…e is instantiated in

### DIFF
--- a/src/Solr/SolrIndex.php
+++ b/src/Solr/SolrIndex.php
@@ -3,7 +3,6 @@
 namespace SilverStripe\FullTextSearch\Solr;
 
 use SilverStripe\Control\Director;
-use SilverStripe\Core\ClassInfo;
 use SilverStripe\Core\Environment;
 use SilverStripe\FulltextSearch\Search\Indexes\SearchIndex;
 use SilverStripe\FullTextSearch\Search\Variants\SearchVariant_Caller;
@@ -1052,12 +1051,9 @@ abstract class SolrIndex extends SearchIndex
      */
     public function uploadConfig($store)
     {
-        // Ensure the short class name is used to match the call which creates the core in SolrService->coreCreate()
-        $indexName = self::getClassNameFromIndex($this->getIndexName());
-        $indexName = ClassInfo::shortName($indexName);
         // Upload the config files for this index
         $store->uploadString(
-            $indexName,
+            $this->getIndexName(),
             'schema.xml',
             (string)$this->generateSchema()
         );
@@ -1065,7 +1061,7 @@ abstract class SolrIndex extends SearchIndex
         // Upload additional files
         foreach (glob($this->getExtrasPath().'/*') as $file) {
             if (is_file($file)) {
-                $store->uploadFile($indexName, $file);
+                $store->uploadFile($this->getIndexName(), $file);
             }
         }
     }
--- a/src/Solr/SolrIndex.php
+++ b/src/Solr/SolrIndex.php
@@ -124,29 +124,6 @@ abstract class SolrIndex extends SearchIndex
         return implode($indexParts);
     }
 
-    /**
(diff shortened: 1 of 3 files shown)


jakedaleweb:
What this was doing was POSTing schema.xml to
`/my/solr/location/silverstripe-namespaced-myindex/`

And the core would then be instantiated in 
`/my/solr/location/myindex/`
Without any of the necessary config files, this would in turn cause the Configure to fail.

dhensby:

dhensby:
@andrewandante tested this locally with the current version of the module and was unable to replicate the described problem...

Has this fix been tested "in the wild" to confirm it is indeed the source of the problem?

jakedaleweb:
@dhensby issue is occurring on a client site and the fix has been tested there.

The work up to this point fixes the Configure task but the Reindex task is now broken (as it relies on the non-namespaced path name) - I am looking in to this now.

jakedaleweb:
Reindex issue has been resolved in `serviceForCore()`

jakedaleweb:
To be clear, at this stage, the module now uses the fully name spaced name across the board for all calls to solr, there were problems with consistency around this earlier where:
- schema.xml etc were being sent to the non-name spaced directory
- the actual cores were instantiated with name spaced dirs
- reindex would attempt to send data to the non name spaced dirs

jakedaleweb:
This branch is now running on mediumtest-uat if you would like to see it in action, the broken version is running on training-uat.

I'm aware there is a broken test now, I'm looking in to it.

src/Solr/SolrIndex.php
```diff
@@ -124,29 +124,6 @@ public function getIndexName()
         return implode($indexParts);
     }
 
-    /**
-     * Helper for returning the indexer class name from an index name, encoded via {@link getIndexName()}
-     *
-     * @param string $indexName
-     * @return string
-     */
-    public static function getClassNameFromIndex($indexName)
```
jakedaleweb: `getIndexName()` in this class serves an almost identical purpose.

tests/SolrIndexTest.php
```diff
@@ -376,30 +376,6 @@ public function testGetIndexNameWithPrefixAndSuffixFromEnvironment()
         );
     }
 
-    /**
-     * @dataProvider indexNameProvider
-     * @param string $indexName
-     * @param string $expected
-     */
-    public function testGetClassNameFromIndex($indexName, $expected)
```
jakedaleweb: Method `getClassNameFromIndex()` was removed, hence the test is also removed

src/Solr/SolrIndex.php
```diff
@@ -1051,17 +1052,20 @@ public function setService(SolrService $service)
      */
     public function uploadConfig($store)
     {
+        // Ensure the short class name is used to match the call which creates the core in SolrService->coreCreate()
+        $indexName = self::getClassNameFromIndex($this->getIndexName());
+        $indexName = ClassInfo::shortName($indexName);
```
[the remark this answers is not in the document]
jakedaleweb: PR has been changed as such

robbieaverill:
Summarising a quick Slack convo with @jakedaleweb about this:

Remaining error: `    "msg": "undefined field _locale",`

FluentSearchVariant (in tractorcow/silverstripe-fluent) applies itself to the environment if the extension is applies, but causes this error on queries when `Locale::getCached()->count();` returns 0.

I fixed this in SS3 for fulltextsearch recently in https://github.com/silverstripe/silverstripe-fulltextsearch/pull/202

Until merge up/for now we can and should also fix this in FluentSearchVariant by also checking that `Locale::getCached()->count() > 0` in appliesToEnvironment

jakedaleweb:
@robbieaverill Referenced PR is here https://github.com/tractorcow/silverstripe-fluent/pull/372/files

This is running on mediumtest-uat, we have successful search there 🙆

PR #208, opened 2018-03-14
NEW Update to use proxied DB instead of self-proxied

### LINKED ISSUE #201: Update to use proxied DB instead of self-proxied
Written up a proxy module at https://github.com/tractorcow/silverstripe-proxy-db. This allows registration of a shared proxy database that multiple modules can decorate.

Also see:
 - https://github.com/lekoala/silverstripe-debugbar/issues/93
 - https://github.com/silverstripe/silverstripe-auditor/issues/16

raissanorth:
This PR makes use of Damian's proxy module https://github.com/tractorcow/silverstripe-proxy-db, which allows the registration of a shared proxy database that multiple modules can decorate.

It fixes #201. 

**Important**: It appears that proxydb can't be used with _any_ other database proxy. This means it relies on https://github.com/silverstripe/silverstripe-auditor/issues/16 being resolved. It currently works with the changes made in https://github.com/lekoala/silverstripe-debugbar/commit/b56a8ed22fa83478246c77522786ff55841d3d52.

Deprecation notices were added in the expectation to introduce these changes in  3.1.0

### DIFF
--- a/_config/database.yml
+++ b/_config/database.yml
@@ -1,32 +1,6 @@
 ---
-Name: fulltextsearchmysql
-After:
-  - 'databaseconnectors'
+Name: fulltextsearchproxydb
 ---
-SilverStripe\Core\Injector\Injector:
-  MySQLDatabase:
-    class: SilverStripe\FullTextSearch\Search\Captures\SearchManipulateCapture_MySQLDatabase
-  MySQLPDODatabase:
-    class: SilverStripe\FullTextSearch\Search\Captures\SearchManipulateCapture_MySQLDatabase
-
----
-Name: fulltextsearchpostgresql
-After:
-  - 'postgresqlconnectors'
----
-SilverStripe\Core\Injector\Injector:
-  PostgrePDODatabase:
-    class: SilverStripe\FullTextSearch\Search\Captures\SearchManipulateCapture_PostgreSQLDatabase
-  PostgreSQLDatabase:
-    class: SilverStripe\FullTextSearch\Search\Captures\SearchManipulateCapture_PostgreSQLDatabase
-
----
-Name: fulltextsearchsqlite
-After:
-  - 'sqlite3connectors'
----
-SilverStripe\Core\Injector\Injector:
-  SQLite3PDODatabase:
-    class: SilverStripe\FullTextSearch\Search\Captures\SearchManipulateCapture_SQLite3Database
-  SQLite3Database:
-    class: SilverStripe\FullTextSearch\Search\Captures\SearchManipulateCapture_SQLite3Database
+TractorCow\SilverStripeProxyDB\ProxyDBFactory:
+  extensions:
+    - SilverStripe\FullTextSearch\Search\Extensions\ProxyDBExtension
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,8 @@
         "silverstripe/framework": "^4.0",
         "monolog/monolog": "~1.15",
         "ptcinc/solr-php-client": "^1.0",
-        "symfony/process": "^3.2"
+        "symfony/process": "^3.2",
+        "tractorcow/silverstripe-proxy-db" : "~0.1"
     },
     "require-dev": {
         "silverstripe/cms": "^4.0",
--- a/src/Search/Captures/SearchManipulateCapture_MySQLDatabase.php
+++ b/src/Search/Captures/SearchManipulateCapture_MySQLDatabase.php
@@ -5,6 +5,10 @@ namespace SilverStripe\FullTextSearch\Search\Captures;
 use SilverStripe\ORM\Connect\MySQLDatabase;
 use SilverStripe\FullTextSearch\Search\Updaters\SearchUpdater;
 
+/**
+ * @deprecated 3.1...4.0 Please use tractorcow/silverstripe-proxy-db to proxy the database connector instead
+ */
+
 class SearchManipulateCapture_MySQLDatabase extends MySQLDatabase
 {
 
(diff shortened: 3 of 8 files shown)


_config/database.yml
```diff
@@ -1,32 +1,7 @@
 ---
-Name: fulltextsearchmysql
-After:
-  - 'databaseconnectors'
+Name: fulltextsearchproxydb
+After: '#fulltextsearchproxydb'
```
robbieaverill: This is likely causing a circular reference to itself - maybe just remove the `After:` line

composer.json
```diff
@@ -24,7 +24,8 @@
         "silverstripe/framework": "^4.0",
         "monolog/monolog": "~1.15",
         "ptcinc/solr-php-client": "^1.0",
-        "symfony/process": "^3.2"
+        "symfony/process": "^3.2",
+        "tractorcow/silverstripe-proxy-db" : "~0.1.0"
```
robbieaverill: I think let's use `~0.1`

src/Search/Captures/SearchManipulateCapture_MySQLDatabase.php
```diff
@@ -5,6 +5,10 @@
 use SilverStripe\ORM\Connect\MySQLDatabase;
 use SilverStripe\FullTextSearch\Search\Updaters\SearchUpdater;
 
+/**
+ * @deprecated since version 3.1.0
```
robbieaverill: Can you add a message suggesting a replacement, e.g. `@deprecated 3.1...4.0 Please use tractorcow/proxydb to proxy the database connector instead`

src/Search/Extensions/ProxyDBExtension.php
```diff
@@ -0,0 +1,19 @@
...
+
+namespace SilverStripe\FullTextSearch\Search\Extensions;
+
+use SilverStripe\Core\Extension;
+use TractorCow\ClassProxy\Generators\ProxyGenerator;
+use SilverStripe\FullTextSearch\Search\Updaters\SearchUpdater;
+
+class ProxyDBExtension extends Extension
+{
+    public function updateProxy(ProxyGenerator &$proxy)
```
robbieaverill: Some doc blocks would be good to explain what this class is doing

raissanorth:
```
/**
 * Class ProxyDBExtension
 * @package SilverStripe\FullTextSearch\Search\Extensions
 *
 * Proxies the database by decorating the factory with code to be extended.
 * Methods can be chained; all addMethod() calls on the same method name will form a set of middleware.
 * First methods registered are executed first.
 *
 */
class ProxyDBExtension extends Extension
{
    /**
     * @param ProxyGenerator $proxy
     */
    public function updateProxy(ProxyGenerator &$proxy)
```
... is a bit to high level, is it?

robbieaverill: I guess so. Devs could get this from the proxy db module.


The point of these manipulateion calls in the context of fulltextsearch is to "Check every index to see what objects need re-inserting into what indexes to keep the index fresh, but doesn't actually do it yet" (see `SearchUpdater::handle_manipulation`.

Perhaps we could summarise a combination of the two... something like "This database connector proxy will allow {@link SearchUpdater::handle_manipulation} to monitor database schema changes that may need to be propagated through to search indexes.".

On a side note, from reading this code I can't actually see where this binds itself to the `Database` class. I see where you're registering the callback for the `manipulate` method, but presumably this doesn't run for every `manipulate` method call in the entire system right? Is there something missing?

robbieaverill:
Also https://github.com/tractorcow/silverstripe-proxy-db/issues/2 looks like it could block this PR?

src/Search/Extensions/ProxyDBExtension.php
raissanorth: @tractorcow Want to help me out with this one, please? I like Robbie's suggestion for the class doc block. When it comes to the updateProxy() one though I find it hard to summarise what I don't fully understand 😉

tractorcow:
@robbieaverill that other issue is caused by the existing proxy on auditor / proxydb. It's not the fault of proxydb.

NightJar:
It does launch a larger question about other proxies outside our control may be rendered incompatible with this new method... But that's not blocking this particular issue/merge.

src/Search/Extensions/ProxyDBExtension.php
raissanorth: Thanks team for your feedback and input!

robbieaverill:
>It does launch a larger question about other proxies outside our control may be rendered incompatible with this new method... But that's not blocking this particular issue/merge.

I think (to be honest) that this is an example of the limitations of using injector to extend things - it works well but only once. Maybe in SS5 there'll be a new way to suggest altering class behaviour... this proxying/middleware approach seems sensible for now

src/Search/Extensions/ProxyDBExtension.php
```diff
@@ -0,0 +1,32 @@
...
+
+namespace SilverStripe\FullTextSearch\Search\Extensions;
+
+use SilverStripe\Core\Extension;
+use TractorCow\ClassProxy\Generators\ProxyGenerator;
+use SilverStripe\FullTextSearch\Search\Updaters\SearchUpdater;
+
+/**
+ * Class ProxyDBExtension
+ * @package SilverStripe\FullTextSearch\Search\Extensions
```
robbieaverill: Can you please remove these two lines? They don't add any value :)

NightJar: iirc `@package` is a deprecated thing now too, right?

PR #215, opened 2018-05-22
[WIP] overhaul docs

andrewandante:
As per numerous requests for clarification and specific examples, this PR aims to improve the documentation provided with the module.

This should also handle CWP-specific configuration to some extent - this is expected to be covered largely by a "shared solr server" section or notes where shared servers differ.


docs/en/03_configuration/31_adding_data_to_an_index.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Publish a page in the CMS
+
+Every change, addition or removal of an indexed class instance triggers an index update through a "processor" object. The update is transparently handled through inspecting every executed database query and checking which database tables are involved in it.
+
+A reindex event will trigger when you make a change in the CMS, via `SearchUpdater::handle_manipulation()`, or `ProxyDBExtension::updateProxy()`. This tracks changes to the database, so any alterations will trigger a reindex. In order to minimise delays to those users, the index update is deferred until after the actual request returns to the user, through PHP's `register_shutdown_function()` functionality.
+
+## Manually
+
+If you get desperate, you can create a new page in a build task or something like that:
```
chillu: Hmm, maybe a bit more professional/confident wording? "desparate" and "something like that" are a bit weak.

docs/en/03_configuration/31_adding_data_to_an_index.md
```diff
@@ -0,0 +1,30 @@
...
+
+```php
+use Page;
+
+$page = Page::create(['Content' => 'Help me. My house is on fire. This is less than optimal.']);
+$page->write();
+```
+
+Depending on the size of the index and how much content needs to be updated, it could take a while for your search results to be updated, so try not to panic if your newly-updated page isn't available immediately.
```
chillu: Same here with "try not to panic", that's a bit patronising.

docs/en/03_configuration/31_adding_data_to_an_index.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Queued jobs
+
+If the queuedjobs module is installed, updates are queued up instead of executed in the same request. Queued jobs are usually processed every minute. Large index updates will be batched into multiple queued jobs to ensure a job can run to completion within common constraints, such as memory and execution time limits. You can check the status of jobs in an administrative interface under `admin/queuedjobs/`.
```
chillu: Link to the module

chillu:
Awesome start, Andrew! :) I think when writing docs which are meant to be read in sequence (the whole "configuration" section?), add a "next section" link to the bottom. Also, consider if some of these pages could be merged. I just ran a poll around README length preferences, and there's a clear preference for longer pages over shorter ones: https://twitter.com/chillu/status/999393971817275392

docs/en/03_configuration/33_dev_tasks.md
```diff
@@ -0,0 +1,19 @@
...
+
+This task will upload configuration to the Solr core, reloading it or creating it as necessary. This should be run after every code change to your indexes, or configuration changes.
+
+## Solr reindex
+
+`dev/tasks/Solr_Reindex`
+
+This task performs a reindex, which adds all the data specified in the index definition into the index store.
+
+If you have the [Queued Jobs module](https://github.com/symbiote/silverstripe-queuedjobs/) installed, then this task will create multiple reindex jobs that are processed asynchronously. Otherwise, it will run in one process. Often, if you are running it via the web, the request will time out. Usually this means the actually process is still running in the background, but it can be alarming to the user.
```
indygriffiths: Might be good to add that it will only do queued jobs when not in dev mode, as per `_config/processor.yml`. A site in dev mode will just index immediately even if queued jobs is installed.

docs/en/03_configuration/30_creating_an_index.md
```diff
@@ -0,0 +1,23 @@
...
+    {
+        $this->addClass(Page::class);
+        $this->addFulltextField('Title');
+    }
+}
+```
+
+This will create a new `SolrIndex` called `MyIndex`, and it will store the `Title` field on all `Pages` for searching.
+
+You can also skip listing all searchable fields, and have the index figure it out automatically via `addAllFulltextFields()`. This will add any database fields that are `instanceof DBString` to the index.
```
indygriffiths: One thing that might be worth adding is that `addAllFulltextFields()` has the potential of indexing sensitive content. Depending on what classes you're indexing and the type of content it contains, explicitly defining the columns to index may be preferred.

docs/en/03_configuration/33_dev_tasks.md
```diff
@@ -0,0 +1,19 @@
...
+
+If instead you run the task via the command line, you will see verbose output as the reindexing progresses.
```
indygriffiths: Worth mentioning the `?verbose=1` trick when running `Solr_Configure` or `Solr_Reindex` from the browser?

andrewandante:
> a clear preference for longer pages over shorter ones

Interesting - so something more like a core page where I have a section, with sub-headings where I have separate pages? Happy to re-arrange it like that.

Style preference is noted, will tighten it up to be more professional.

Thanks for all your feedback so far @chillu and @indygriffiths !

andrewandante:
Changed my mind and went with the re-structure first 😅

chillu:
> I also realise that I'm writing these from the perspective of assuming it's connected to Solr - can we safely assume that if someone is using this module, they are using Solr?

Sorry just saw this. It's a abstraction layer in theory rather than in practice, and I don't see us adding Sphinxsearch or ElasticSearch to this layer anytime soon. If we decide to add/change our hosting from Solr to ElasticSearch, I think it would be a good point to reconsider the scope of what this module does in the first place, so I doubt we'd be adding it to this particular API surface. So long story short, it's fine to heavily gear the docs towards Solr use.

andrewandante:
Squashed up what I've done, but this will need to be picked up by someone else now most likely - @dhensby leading candidate 👍

dhensby:
I'm happy with where this is. Unless we have a list of outstanding docs to write, this looks good as is :)

any new docs can be added in another PR

chillu:
Created a follow up ticket to harmonise this with CWP docs https://github.com/silverstripe/cwp/issues/126

bin/fts_quickstart
```diff
@@ -0,0 +1,87 @@
...
+echo "Installing Java SDK..."
+# Find our package manager
+if VERB="$( command -v apt-get )" 2> /dev/null; then
+   echo "Debian-based OS detected"
+   sudo apt-get install -y openjdk-8-jdk 2> /dev/null
+elif VERB="$( command -v yum )" 2> /dev/null; then
+   echo "Modern Red Hat-based OS detected"
+   sudo yum install -y java-1.8.0-openjdk.x86_64 2> /dev/null
+else
+   echo "No valid package manager detected; try one of apt-get, yum."
```
Firesphere: What about pacman or SuSe's YaST? (Just checking ;) )

PR #216, opened 2018-06-04
Replace default CMS SearchForm with Solr version

andrewandante:
Assume that if we have installed the module, we want the `$SearchForm` to use Solr, not the `FulltextSearch::enable()` version.

Will add docs shortly - and this is a very basic implementation, to be used as a jumping off point mostly I think.

### DIFF
--- a/_config/config.yml
+++ b/_config/config.yml
@@ -4,3 +4,6 @@ Name: fulltextsearchconfig
 SilverStripe\ORM\DataObject:
   extensions:
     - SilverStripe\FullTextSearch\Search\Extensions\SearchUpdater_ObjectHandler
+SilverStripe\CMS\Controllers\ContentController:
+  extensions:
+    - SilverStripe\FullTextSearch\Solr\Control\ContentControllerExtension
--- a/docs/en/index.md
+++ b/docs/en/index.md
@@ -46,7 +46,11 @@ SilverStripe\FullTextSearch\Search\Updaters\SearchUpdater:
 
 ## Basic usage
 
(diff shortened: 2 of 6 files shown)


andrewandante:
Considering adding a better template that loops over the Result data, rather than the Matches data, so that it can be better expanded on. 🤔

andrewandante:
Have instead added a `Page_results_solr` template and returned the `Results` instead of just `Matches` for better extensibility.

NightJar:
I don't think it's so much to _replace_ the default search with the Solr backed one, but rather that one can easily use it _instead_ of the default (`silverstripe/framework` [`SilverStripe\ORM\Search\Fulltextsearchable::enable`](https://github.com/silverstripe/silverstripe-framework/blob/4/src/ORM/Search/FulltextSearchable.php#L81)) - in that if both are applied there will be conflicts about which form is used, no?

Perhaps making this clear by updating the Readme (and/)or some other documentation to point this out would be nice too, providing my assumption is correct :)

andrewandante:
@NightJar fair point - if both are applied then I'm not sure which will apply, they are both extensions after all. 

Will update the README to be more explicit about that 👍

andrewandante:
Added a line in the README - will update even more once merged and I can get at it via #215

NightJar:
Hmm, I would suggest an `After` directive in the config header, but `FulltextSearchable` applying via `php` makes it _always_ the last thing to be processed. So that's not helpful :)
Nice work, time to mergey :D

andrewandante:
Thanks @NightJar - you da best 😍

PR #217, opened 2018-06-12
add quickstart script into vendor/bin

andrewandante:
Pulled out from #215 

Simplifies installation on a (yum- , dnf- or apt-ready) dev server to be:

```bash
composer require silverstripe/fulltextsearch && vendor/bin/fulltextsearch_quickstart
```

### DIFF
--- a/bin/fulltextsearch_quickstart
+++ b/bin/fulltextsearch_quickstart
@@ -1,0 +1,94 @@
+#!/usr/bin/env bash
+
+echo "[WARNING] This is intended for development environments only."
+echo "Do NOT use in Production."
+
+echo "Installing Java SDK..."
+# Find our package manager
+if VERB="$( command -v apt-get )" 2> /dev/null; then
+   echo "Debian-based OS detected"
+   sudo apt-get install -y openjdk-8-jdk 2> /dev/null
+elif VERB="$( command -v dnf )" 2> /dev/null; then
+   echo "Modern Fedora-based OS detected"
+   sudo dnf install -y java-1.8.0-openjdk.x86_64 2> /dev/null
+elif VERB="$( command -v yum )" 2> /dev/null; then
+   echo "Fedora-based OS detected"
+   sudo yum install -y java-1.8.0-openjdk.x86_64 2> /dev/null
+else
+   echo "No valid package manager detected; try installing one of the following:"
+   echo "apt-get, dnf, yum"
+   exit 1
+fi
+
+if [ ! -d "/opt/solr" ]; then
+    printf "Installing Solr 4"
+    # Acquire and unzip solr4
+    wget http://archive.apache.org/dist/lucene/solr/4.10.4/solr-4.10.4.tgz 2> /dev/null && printf "."
+    tar -xf solr-4.10.4.tgz 2> /dev/null && printf "."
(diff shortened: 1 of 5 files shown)


bin/fts_quickstart
```diff
@@ -0,0 +1,87 @@
...
+# Get solr running
+sudo /etc/init.d/solr start 2> /dev/null
+
+# Determine application dir
+if [ -d app ]; then
+    APPDIR="app"
+elif [ -d mysite ]; then
+    APPDIR="mysite"
+else
+    echo "Can't detect application dir - skipping default index creating"
```
dhensby:
```diff
-echo "Can't detect application dir - skipping default index creating"
+echo "Can't detect application dir - skipping default index creation"
```

bin/fts_quickstart
```diff
@@ -0,0 +1,87 @@
...
+    echo "]);" >> "$APPDIR/_config.php"
+fi
+
+# Determine code dir
+if [ -d "$APPDIR/src" ]; then
+    CODEDIR="$APPDIR/src"
+elif [ -d "$APPDIR/code" ]; then
+    CODEDIR="$APPDIR/code"
+else
+    echo "Can't detect code dir - skipping default index creating"
```
dhensby:
```diff
-echo "Can't detect application dir - skipping default index creating"
+echo "Can't detect application dir - skipping default index creation"
```

bin/fts_quickstart
```diff
@@ -0,0 +1,87 @@
...
+    APPDIR="app"
+elif [ -d mysite ]; then
+    APPDIR="mysite"
+else
+    echo "Can't detect application dir - skipping default index creating"
+    exit 1
+fi
+
+# Check to see if it has been configured in _config.php
+grep -i "Solr::configure_server(" "$APPDIR/_config.php" 2> /dev/null
```
dhensby: you can use the `-q` argument with grep to suppress output and only give you an exit code

bin/fts_quickstart
```diff
@@ -0,0 +1,87 @@
+#!/usr/bin/env bash
+
+echo "Installing Java SDK..."
+# Find our package manager
+if VERB="$( command -v apt-get )" 2> /dev/null; then
+   echo "Debian-based OS detected"
+   sudo apt-get install -y openjdk-8-jdk 2> /dev/null
+elif VERB="$( command -v yum )" 2> /dev/null; then
```
dhensby: `dnf` is preferred in fedora now, you should try `dnf` first, then `yum`

bin/fts_quickstart
```diff
@@ -0,0 +1,87 @@
...
+   echo "Modern Red Hat-based OS detected"
+   sudo yum install -y java-1.8.0-openjdk.x86_64 2> /dev/null
+else
+   echo "No valid package manager detected; try one of apt-get, yum."
```
dhensby: Seems a bit silly to recommend trying package managers you've already tried.

andrewandante: The intent was to recommend to the dev to try installing one of `apt` or `yum` - I can update the message to be clearer.

dhensby: Ah, I see! yes, clearer is better :)

PR #223, opened 2018-06-27
FIX Use \Exception for catching Solr exceptions

ScopeyNZ:
As titled. Existing try...catch blocks are not catching errors that are intended to be caught.


robbieaverill:
Merged up, thanks!

PR #224, opened 2018-06-27
FIX Reduce log level so errors do not automatically output

### LINKED ISSUE #222: Incorrectly configured Solr causes uncaught exception fatal errors
Currently when attempting to add new CMS pages, or when trying to save existing CMS pages with incorrectly configured Solr you will get uncaught exceptions causing fatal errors. For example:

image

In SS3 this was not an issue.

ScopeyNZ:
Resolves #222 

This reduces the log level. Unfortunately the method name is still `warn` which could be considered inconsistent but I didn't want to change the API.

### DIFF
--- a/src/Search/Indexes/SearchIndex.php
+++ b/src/Search/Indexes/SearchIndex.php
@@ -565,7 +565,7 @@ abstract class SearchIndex extends ViewableData
      */
     public static function warn($e)
     {
-        Injector::inst()->get(LoggerInterface::class)->warning($e);
+        Injector::inst()->get(LoggerInterface::class)->info($e);
     }
 
     /**


src/Search/Indexes/SearchIndex.php
```diff
@@ -565,7 +565,7 @@ protected function _getFieldValue($object, $field)
      */
     public static function warn($e)
```
NightJar: You could change the method name to `info` as well, so long as you also implement a `warn` function that marks deprecated and calls `info` :)
It would require a minor number version bump though, but I think that's probably fine.

ScopeyNZ: Good idea. I could add a new method but ideally this would be best as a patch so it can stop breaking things. Currently these errors are just dumped out regardless of context - especially annoying in `Content-Type: application/json` responses...

ScopeyNZ: Haha. I think eventually - perhaps when the default logger implementation is improved in core, there could be some work done here. Switch it back to a more appropriate level, preferably remove the random public API on this class for logging. Considering that, I'd prefer to not do any API updates, as I think the API could be improved significantly with a more directed PR.

NightJar: OK, cool :)

Issue #228, opened 2018-08-27
Pagination in search results broken


sanderha:
I might be missing something, but it seems generated search results does not care about the "start" GET variable.
Shouldnt this method `SilverStripe\FullTextSearch\Solr\Forms\SearchForm::getResults()` account for that?
I fixed it in my own `SearchForm` class by adding this:
```
        $searchTerms = $request->requestVar('Search');
        $start = $request->requestVar('start');
        $query = SearchQuery::create()->addSearchTerm($searchTerms)->setStart($start);
```

Apologies if I missed something obvious

Using SS v4.1 and v3.3 of this module

robbieaverill:
@sanderha would you like to make a pull request to fix this?

sanderha:
#229 There ya go

robbieaverill:
Fixed in #229, thank you

PR #229, opened 2018-08-28
fix missing pagination in search results

sanderha:
Results in SearchForm did not take start GET var into account

### DIFF
--- a/src/Solr/Forms/SearchForm.php
+++ b/src/Solr/Forms/SearchForm.php
@@ -63,7 +63,10 @@ class SearchForm extends Form
         $request = $this->getRequestHandler()->getRequest();
 
         $searchTerms = $request->requestVar('Search');
-        $query = SearchQuery::create()->addSearchTerm($searchTerms);
+        // for pagination
+        $start = $request->requestVar('start');
+
+        $query = SearchQuery::create()->addSearchTerm($searchTerms)->setStart($start);
 
         $params = [
             'spellcheck' => 'true',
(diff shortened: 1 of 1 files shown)


sanderha:
#228

src/Solr/Forms/SearchForm.php
```diff
@@ -63,7 +63,10 @@ public function getResults()
         $request = $this->getRequestHandler()->getRequest();
 
         $searchTerms = $request->requestVar('Search');
-        $query = SearchQuery::create()->addSearchTerm($searchTerms);
+        // for pagination
+        $start = $request->requestVar('start');
```
robbieaverill: Perhaps we should check whether it has a value before setting it, that way we still get the default value if no argument is provided

sanderha:
@robbieaverill added a conditional to check for the start GET variable first

src/Solr/Forms/SearchForm.php
```diff
@@ -65,6 +65,10 @@ public function getResults()
         $searchTerms = $request->requestVar('Search');
         $query = SearchQuery::create()->addSearchTerm($searchTerms);
 
+        if($start = $request->requestVar('start')){
```
robbieaverill: Just PSR-2 compliance to go, fixing now for you

robbieaverill:

robbieaverill:
Thank you!

sanderha:
Ah yeah, sorry about that. Will think about it next time. The more you know :-)

PR #234, opened 2018-10-24
Update Process timeout via Config

Leapfrognz:
Currently SolrReindexImmediateHandler will die after 60 seconds, as that is the default timeout in Symfony\Component\Process\Process. This will allow that to be overridden via Config

### DIFF
--- a/src/Solr/Reindex/Handlers/SolrReindexImmediateHandler.php
+++ b/src/Solr/Reindex/Handlers/SolrReindexImmediateHandler.php
@@ -98,11 +98,8 @@ class SolrReindexImmediateHandler extends SolrReindexBase
         // Execute script via shell
         $process = new Process($cmd);
 
-        // Set timeout from config. Process default is 60 seconds.
-        $timeout = Config::inst()->get(static::class, 'process_timeout');
-        if ($timeout) {
-            $process->setTimeout($timeout);
-        }
+        // Set timeout from config. Process default is 60 seconds
+        $process->setTimeout($this->config()->get('process_timeout'));
 
         $process->inheritEnvironmentVariables();
         $process->run();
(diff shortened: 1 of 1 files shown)


NightJar:
Thanks @Leapfrognz - are you able to confirm that this can be disabled via a config setting of `null`?
Would you mind writing a test for this code?

Leapfrognz:
@Dylan, Id love to write a unit test. Can you wait a year? :s
Tested locally, leaving out the yml setting entirely or setting it null defaults to Process.php's 60 second timeout

src/Solr/Reindex/Handlers/SolrReindexImmediateHandler.php
```diff
@@ -97,6 +97,13 @@ protected function processGroup(
 
         // Execute script via shell
         $process = new Process($cmd);
+
+        // Set timeout from config. Process default is 60 seconds.
+        $timeout = Config::inst()->get(static::class, 'process_timeout');
```
robbieaverill: Prefer `$this->config()->get(‘process_timeout’)`

src/Solr/Reindex/Handlers/SolrReindexImmediateHandler.php
```diff
@@ -97,6 +97,13 @@ protected function processGroup(
...
+        if ($timeout) {
```
robbieaverill: Yeah, as @NightJar said, null is a valid value to pass through. This if statement could be removed entirely

NightJar:
Neat, thanks James.
Personally I'm OK with it being entirely disabled by default, ideally it always should have been (I don't understand Symfony's default).

Leapfrognz:
Thanks Guys! Quick response is appreciated.

Leapfrognz:
@NightJar 
Actually `$this->Config()->get('process_timeout')` doesnt work. Can we change it back to `Config::inst(self::class)->get('process_timeout')` ?

robbieaverill:
Ah you need to add the Configurable trait to `SolrReindexBase`

PR #235, opened 2018-10-30
FIX: Use Configurable trait

Leapfrognz:


robbieaverill:
It would probably be tidier to add it to `SolrReindexBase`, but it's OK for now

PR #242, opened 2018-12-04
FIX #241 - Support numerically indexed array of classes

### LINKED ISSUE #241: SearchVariant::withCommon expects classes in an array indexed by class names
The newly introduced SearchVariant::withCommon method expects a list of classes in an array indexed by class names, which breaks the previously functioning behaviour with just numeric keys. The class is already expected in the 'class' option anyway, so it's an unnecessary duplication.

michalkleiner:
Re-add the option to use numerically indexed arrays of classes in `withCommon` method, similar to what `with` method allowed.

### DIFF
--- a/src/Search/Variants/SearchVariant.php
+++ b/src/Search/Variants/SearchVariant.php
@@ -187,8 +187,8 @@ abstract class SearchVariant
         $commonVariants = [];
         foreach ($classes as $class => $options) {
             // BC for numerically indexed list of classes
-            if (is_numeric($class) && isset($options['class'])) {
-                $class = $options['class'];
+            if (is_numeric($class) && !empty($options['class'])) {
+                $class = $options['class']; // $options['class'] is assumed to exist throughout the code base
             }
 
             // Extract relevant class options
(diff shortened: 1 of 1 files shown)


michalkleiner:
@robbieaverill what's the best target version for this to get in? Possibly `3`?

NightJar:
The lowest compatible version that first presented the bug, generally @michalkleiner :)

src/Search/Variants/SearchVariant.php
```diff
@@ -186,6 +186,11 @@ public static function withCommon(array $classes = [])
         // Construct new array of variants applicable to at least one class in the list
         $commonVariants = [];
         foreach ($classes as $class => $options) {
+            // BC for numerically indexed list of classes
+            if (is_numeric($class) && isset($options['class'])) {
+                $class = $options['class'];
```
NightJar: Would you think it reasonable to throw an exception or error here if `$options['class']` is unset?
Or alternatively, add a comment to indicate that it's _always_ set?

michalkleiner: Other pieces of code in the module rely on the `$options['class']` and they won't work without it, so I'd say it's nearly safe to assume it exists. Someone can make the decision and I'll update it.

NightJar: I can make the decision - just add a comment reflecting this info and I'll merge it now* :)

e.g.
```diff
+ // $options['class'] is always assumed to exist throughout the code base
```

*after the test suite passes

PR #246, opened 2019-03-24
Adds a 'post' option for solrconfigstore

jakedaleweb:
# What 
This copies the `CwpSolrConfigStore` ([from here](https://github.com/silverstripe/cwp-search/blob/master/src/Solr/CwpSolrConfigStore.php)) in to the fulltextsearch module. If merged that file can be removed from the cwp-search module and the mode set to 'post' in `CwpSolr.php`

# Why
The CWP solrproxy which receives the solr configuration files via POST is going to be used on SSP also. This requires the application to configure a config store which does a POST to send through the files. As this config store is not strongly tied to the CWP solr proxy, just an endpoint that requires a POST I think it is generic enough to put in this module. This will mean users of solr on SSP will not need to install CWP modules to use solr.

### DIFF
--- a/src/Solr/Solr.php
+++ b/src/Solr/Solr.php
@@ -32,13 +32,19 @@ class Solr
      *
      * indexstore => an array with
      *
-     *   mode - a classname which implements SolrConfigStore, or 'file' or 'webdav'
+     *   mode - a classname which implements SolrConfigStore, or 'file', 'webdav' or 'post'
      *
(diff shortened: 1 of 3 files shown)


robbieaverill:
In terms of your suggestion for implementing this into CWP, we'll need to deprecate the previous CWP config store in a minor release (e.g. the master branch of cwp/cwp-search) and re-configure CWP to use this instead. We probably won't be doing another major release of cwp/cwp-search, but deprecating is fine for now.

Nice work, thanks for the PR!

PR #255, opened 2019-05-30
Disable indexing on file migration

maxime-rainville:
This PR disables indexing when the file migration is doing its thing.

It shouldn't break if the MigrateFIleTask is not at it's expected spot or if the extension points hasn't been implemented in the project's framework version.

(I've never done anything to fulltextsearch before, so don't assume I know what I'm doing, because I don't.)

# Parent issue
* https://github.com/silverstripe/silverstripe-framework/issues/8965


NightJar:
Ideal, nice work 👍

PR #256, opened 2019-05-31
Clarified quick start steps vs manual config steps when requiring

taoceanz:
vendor/bin/fulltextsearch_quickstart attempts to copy two files (daemon and default_index.php.example) from the docs folder, except .gitattributes file prevents the docs folder from being required when requiring via composer.
`sudo cp vendor/silverstripe/fulltextsearch/docs/examples/daemon_script /etc/init.d/solr 2> /dev/null`
`cp vendor/silverstripe/fulltextsearch/docs/examples/default_index.php.example "$CODEDIR/FulltextSearch/DefaultIndex.php"`

I'd imagine it better to pull those two files from the github repo in the fulltestsearch_quickstart script and then revert this 01_getting_started.md doc, though I'm unsure exactly how to require the exact project path to ensure the correct version file is retrieved for the module version.
`wget -O - https://raw.githubusercontent.com/<username>/<project>/<branch>/<path>/<file> | bash`

### DIFF
--- a/docs/en/01_getting_started.md
+++ b/docs/en/01_getting_started.md
@@ -22,24 +22,32 @@ fulltext searching as an extension of the object model. However, the disconnect 
 design and the object model meant that searching was inefficient. The abstraction would also often break and it was
 hard to then figure out what was going on.
 
-This module instead provides the ability to define those indexes and queries in PHP. The indexes are defined as a 
-mapping between the SilverStripe object model and the connector-specific fulltext engine index model. This module then 
+This module instead provides the ability to define those indexes and queries in PHP. The indexes are defined as a
+mapping between the SilverStripe object model and the connector-specific fulltext engine index model. This module then
 interrogates model metadata to build the specific index definition.
 
-It also hooks into SilverStripe framework in order to update the indexes when the models change and connectors then 
+It also hooks into SilverStripe framework in order to update the indexes when the models change and connectors then
 convert those index and query definitions into fulltext engine specific code.
 
 The intent of this module is not to make changing fulltext search engines seamless. Where possible this module provides
 common interfaces to fulltext engine functionality, abstracting out common behaviour. However, each connector also
-offers its own extensions, and there is some behaviour (such as getting the fulltext search engines installed, 
+offers its own extensions, and there is some behaviour (such as getting the fulltext search engines installed,
 configured and running) that each connector deals with itself, in a way best suited to that search engine's design.
 
 ## Quick start
 
+If you intend to get up and running without using the fulltextsearch_quickstart script, require this module via composer:
+
+```bash
+composer require silverstripe/fulltextsearch
+```
+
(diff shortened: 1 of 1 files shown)


docs/en/01_getting_started.md
```diff
@@ -22,24 +22,32 @@ fulltext searching as an extension of the object model. However, the disconnect
...
+composer require silverstripe/fulltextsearch
+```
+
+You're now ready to begin [set up](01_setup.md).
+
 If you are running on a Linux-based system, you can get up and running quickly with the quickstart script, like so:
 
 ```bash
-composer require silverstripe/fulltextsearch && vendor/bin/fts_quickstart
+git clone https://github.com/silverstripe/silverstripe-fulltextsearch.git vendor/silverstripe/fulltextsearch && vendor/bin/fulltextsearch_quickstart
```
robbieaverill: I'm not sure I agree with this change. Is there a reason you're suggesting against using Composer?

taoceanz: Using composer to require the module doesn't require the docs folder, but the fulltextsearch_quickstart script tries to copy files from the docs/examples folder on line 38 and line 88.
Without those files, the script doesn't start Solr and doesn't create a default index.
.gitattributes prevents the docs folder being exported.

In vendor/bin/fulltextsearch_quickstart:
Line 38 copies daemon_script into /etc/init.d/solr before solr is started on line 44.
Line 88 copies default_index.php.example into $CODEDIR/FulltextSearch/DefaultIndex.php.

robbieaverill: Ah I see. https://github.com/silverstripe/silverstripe-fulltextsearch/pull/255 fixed that by suggesting to install from source instead. We need to merge that up to master branch which I will do now, but perhaps you could remove this part of the change in lieu of that

taoceanz: Ah, yeah `--prefer-source` is far more elegant! Thanks @robbieaverill

robbieaverill:
Thanks for the PR anyway, I've merged up the branches to master. Were there any other parts of this PR you'd like to contribute separately?

taoceanz:
No other parts. Everything there was to address requiring the docs file which --prefer-source addresses much more succinct and elegantly.

taoceanz:
Requiring with composer doesn't require the docs folder which the
fulltextsearch_quickstart script refers to to copy files into another
location.
.gitattributes prevents the docs folder from being required by composer.
The script needs the daemons and example index file from docs/example.

robbieaverill:
We shouldn’t have docs in gitattributes IMO

taoceanz:
I agree docs shouldn't be in gitattributes, but didn't want to step on toes removing it.

robbieaverill:
Go for gold!

taoceanz:
Your wish is my command, master.

#258

PR #258, opened 2019-05-31
Make docs available without --prefer-source

taoceanz:
Make docs available in composer require without preferring source to ensure vendor/bin/fulltextsearch_quickstart script has access to the files it needs and users have docs available available locally.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,4 @@
 /tests export-ignore
-/docs export-ignore
 /.gitattributes export-ignore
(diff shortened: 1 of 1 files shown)


robbieaverill:
It's worth noting that this won't help the original problem until the next tagged release. You can install dev branches locally if you want, which won't exclude things listed in .gitattributes, e.g. `3.0.x-dev` rather than `^3.0` in composer dependency constraints

PR #259, opened 2019-06-13
 Define getRunAsMemberID

wilr:
On the latest v4.4 CWP pulling down the queued jobs looks like there needs to be an additional definition on each job.

https://github.com/symbiote/silverstripe-queuedjobs/commit/d6160257f74164864454d7766533e946a9d3b4c5

### DIFF
--- a/src/Search/Processors/SearchUpdateCommitJobProcessor.php
+++ b/src/Search/Processors/SearchUpdateCommitJobProcessor.php
@@ -273,4 +273,9 @@ class SearchUpdateCommitJobProcessor implements QueuedJob
     {
         return $this->messages;
     }
+    
+    public function getRunAsMemberID()
+    {
+        return 0;
+    }
 }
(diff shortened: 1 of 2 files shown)


NightJar:
This is technically a minor release change, could you please re-target the PR to the `3.6`* branch?


*which kinda weirdly already exists, although there's no `3.6.0` release yet it seems... :confused:

wilr:
@NightJar I targeted 3.5 since that is currently what CWP 2.3 pulls in and CWP 2.3 pulls in the queued job change.

robbieaverill:
@wilr was there a breaking change in FTS? We should try and fix it there if we can

robbieaverill:
This is the relevant change FYI https://github.com/symbiote/silverstripe-queuedjobs/pull/247/files#diff-bf97514c80202ef07479438c4f389368R213

wilr:
@robbieaverill only other recommendation rather than updating all the queued jobs (which didn't extend AbstractQueuedJob) ever written is to remove the method from the Interface on queued jobs and just use a `hasMethod`

robbieaverill:
Yeah, to be fair I don't think putting it in the interface was a safe change to make in a minor release anyway.

emteknetnz:
Looks like the breaking change caused by the introduction of a new interface method was resolved by creating a new interface here https://github.com/symbiote/silverstripe-queuedjobs/pull/262/files

emteknetnz:
Closing as we no longer appear to require this pull-request.  Thank you for originally submitting it.

PR #266, opened 2019-08-26
Added translation support for results page + Finnish translation.

mediaclinic:


templates/Layout/Page_results_solr.ss
```diff
@@ -2,11 +2,11 @@
     <h1>$Title</h1>
 
     <% if $Query %>
-        <p class="searchQuery">You searched for &quot;{$Query}&quot;</p>
+        <p class="searchQuery"><%t Results.SearchQuery 'You searched for' %> &quot;{$Query}&quot;</p>
```
robbieaverill: `Results.SearchQuery` is a global translation key, so it's not very unique. Perhaps you could use `SolrResultsPage.SearchQuery` instead?

robbieaverill:

templates/Layout/Page_results_solr.ss
mediaclinic: Thanks, I changed the key to be more unique.

templates/Layout/Page_results_solr.ss
```diff
@@ -23,34 +23,34 @@
                         </a>
                     </h4>
                     <p><% if $Abstract %>$Abstract.XML<% else %>$Content.ContextSummary<% end_if %></p>
-                    <a class="readMoreLink" href="$Link" title="Read more about &quot;{$Title}&quot;">Read more about &quot;{$Title}&quot;...</a>
+                    <a class="readMoreLink" href="$Link" title="Read more about &quot;{$Title}&quot;"><%t SolrResultsPage.ReadMore 'Read more about' %> &quot;{$Title}&quot;...</a>
```
robbieaverill: the title here isn’t translated - since it’s the same as the content, it doesn’t provide any extra value, maybe you could remove the attribute

mediaclinic: Maybe I just add translation to title also, as other links uses title attribute?

robbieaverill: You can, but it provides no value when the title is the same as the content, so may as well be removed. Either way this is an improvement, thank you!

PR #271, opened 2020-01-07
FIX: #270 - Reset variant state back to original state after processing

### LINKED ISSUE #270: Variant state not reset during processing.
The `SolrReindexBase` class groups changes by Variants and processes them in batches. These Variants (E.g. StageVariant, SubsiteVariant, etc) are used to ensure content is only indexed against a specific state. (E.g. Live content on Subsite 3).

`SolrReindexBase` sets the state accordingly when processing, however it does not reset you back to your original state.

https://github.com/silverstripe/silverstripe-fulltextsearch/blob/afc6364d22d2d2f7d811d791b0bfdb24c677930f/src/Solr/Reindex/Handlers/SolrReindexBase.php#L171

This causes issues when you have mutliple indexes that uses different Variants as the current state is not reset causing the Variants to also be applied to the next set of changes (unless it is specifically overriden).

E.g.

Index1 = (VariantStage + VariantSubsite)
Index2 = (VariantStage)

Once Index1 has finished processing, it does not reset the Subsite state, causing Index2 to also apply the VariantSubsite that was last used (E.g. the last SubsiteID that was processed).

PR for fix raised - #271

brettt89:
Fixes #270 

 - Reset state to originalState when processing Variants / Groups

### DIFF
--- a/src/Solr/Reindex/Handlers/SolrReindexBase.php
+++ b/src/Solr/Reindex/Handlers/SolrReindexBase.php
@@ -99,6 +99,8 @@ abstract class SolrReindexBase implements SolrReindexHandler
         $batchSize,
         $taskName
     ) {
+        // Get current state
+        $originalState = SearchVariant::current_state();
         // Set state
         SearchVariant::activate_state($state);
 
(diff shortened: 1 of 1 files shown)


robbieaverill:
You may want to retarget this PR at a minor release branch :-)

brettt89:
Hmm I think my rebasing may have failed a bit here. One moment caller.

brettt89:
@robbieaverill - Done :) . Ended up re-creating branch from 3.6 and cherry picking the commit. Need to brush up on my git skills again I think haha.

robbieaverill:
Nice, that’s essentially what a rebase is anyway!

PR #272, opened 2020-02-10
FIX: SearchUpdateQueuedJobProcessor JobType NULL

adrexia:
getJobType method returns null as querying a string rather than a namespaced class.


adrexia:
Yeah, true. I was just going for minimal change, but will update

PR #273, opened 2020-02-10
FIX: allow update processor to be overridden by existing yml

adrexia:
One of the 3->4 update commits (6066af5841fd1df21ef32ff3fa089c6e48ea4f26) replaced `SearchUpdateProcessor` with `SearchUpdateImmediateProcessor`, but that means that the existing processor yml fails to replace the `SearchUpdateImmediateProcessor` with `SearchUpdateQueuedJobProcessor`  as I think is intended?

c.f https://github.com/silverstripe/silverstripe-fulltextsearch/blob/3/_config/processor.yml#L19

### DIFF
--- a/src/Search/Updaters/SearchUpdater.php
+++ b/src/Search/Updaters/SearchUpdater.php
@@ -11,7 +11,8 @@ use SilverStripe\ORM\DB;
 use SilverStripe\FullTextSearch\Search\FullTextSearch;
 use SilverStripe\FullTextSearch\Search\SearchIntrospection;
 use SilverStripe\FullTextSearch\Search\Variants\SearchVariant;
-use SilverStripe\FullTextSearch\Search\Processors\SearchUpdateImmediateProcessor;
+use SilverStripe\FullTextSearch\Search\Processors\SearchUpdateProcessor;
+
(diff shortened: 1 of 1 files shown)


emteknetnz:
Thanks again for creating this pull-request.

Before this gets merged, the team at Silverstripe Ltd are going to need to perform some testing on this work, though at this stage we're not sure when this will get prioritised.

If someone else in the community is able to take over testing duties for this one so that this can be merged sooner, then that would be certainly appreciated.

During the testing phase we may make some alterations to the pull-request, though rest-assured, I'll ensure that you still get credited for the work you've done in the release changelog.

chillu:
I've tested this locally on `cwp-installer:2.5.x-dev` and it works as expected. Thanks for digging into this Naomi. As discussed on the [Helpdesk ticket](https://silverstripe.freshdesk.com/a/tickets/61921), we also need to look at the autoSoftCommit settings.

adrexia:
@chillu thanks!

RE: autoSoftCommit  - there is discussion on this ticket that needs a resolution/direction: https://github.com/silverstripe/silverstripe-fulltextsearch/issues/274

Issue #257, opened 2019-05-31
Don't index records with ShowInSearch=false


chillu:
The module makes no assumption about what criteria might be applied for records to be added to the index. I'm proposing elsewhere that we [filter out records with anonymous canView=false](https://github.com/silverstripe/silverstripe-fulltextsearch/issues/252). 

In addition, I think we should add behavior to prevent indexing of a record based on custom, model specific flags like `ShowInSearch`. Currently, the docs mention this usage pattern (e.g. https://www.cwp.govt.nz/developer-docs/en/2/features/solr_search/default_search_index/), and recommend adding this as a filter field. You're still in charge of filtering on that field by default. The [CWPSearchEngine](https://github.com/silverstripe/cwp-search/blob/master/src/CwpSearchEngine.php#L61) performs this filter out of the box, but *only* on `SiteTree` and `File` (see [pull request](https://github.com/silverstripe/cwp/pull/64/files#r178983597)). If you're defining your own index on other objects, or include other objects in that index, you need to create your own search engine - the filter field is indexed, but it's ignored during searches. The docs for [searching dataobjects](https://www.cwp.govt.nz/developer-docs/en/2/features/solr_search/searching_dataobjects/) fail to mention that as well. This illustrates why we shouldn't have those records in the index in the first place.

This has some of the same rollout issues as [filter out records with anonymous canView=false](https://github.com/silverstripe/silverstripe-fulltextsearch/issues/252), we'd need to advise devs on how to get those existing index entries with ShowInSearch=false out of their index.

When releasing this, we should strongly recommend to search users that they run a reindex to prune any records with ShowInSearch=0 in their current search state.

Another thing to check: when we change ShowInSearch from 1 to 0 and click save/publish, does it get removed from the index straight away?

### ACs
- [x] Records which have a ShowInSearch method or database property returning false are not indexed
- [x] This applies to built-in records (Page, File) as well as custom record definitions
- [x] This behaviour applies to both CWP and core module search
- [x] The ShowInSearch value is inspected every time a record is saved (non-versioned) or published (versioned), and the index entry is updated accordingly
- [x] Existing index entries with ShowInSearch=0 can be pruned through a reindex
- [x] There is documentation both for CWP and generic module upgrade paths
- [x] Documentation is adjusted to state that you MUST filter out ShowInSearch during search request for safety (rather than reindexing) Steve: users don't need to filter out during search, happens automatically 
[...]

PR #275, opened 2020-03-26
Filter DataObjects on ShowInSearch

emteknetnz:
**This was originally branched off 3.6, have since rebased and retargeted to 3**

Issue: https://github.com/silverstripe/silverstripe-fulltextsearch/issues/257

**Inlet filtering - cms save**
SearchUpdateProcessor.php

**Inlet filtering - solr reindex**
SolrReindexBase.php 

**Outlet filtering**
SolrIndex.php

### DIFF
--- a/docs/en/03_configuration.md
+++ b/docs/en/03_configuration.md
@@ -147,7 +147,7 @@ to render properly in the search results:
 * `Link()` needs to return the URL to follow from the search results to actually view the object.
 * `Name` (as a DB field) will be used as the result title.
 * `Abstract` (as a DB field) will show under the search result title.
-* `getShowInSearch()` is required to get the record to show in search, since all results are filtered by `ShowInSearch`.
+* `ShowInSearch` (as a DB field) or `getShowInSearch()` is recommended to allow the optional exclusion of DataObjects from being added to the search index.  If omitted, then all DataObjects of this type will be added to the search index.
 
 So with that, you can add your class to your index:
 
@@ -171,6 +171,28 @@ Once you've created the above classes and run the [solr dev tasks](#solr-dev-tas
 you've just created, this will add `SearchableDataObject` and the text fields it has to the index. Now when you search 
 on the site using `MySolrSearchIndex->search()`, the `SearchableDataObject` results will show alongside normal `Page`
 results.
+
+### ShowInSearch and getShowInSearch() filtering
+
+The fulltextsearch module checks the value of `ShowInSearch` on each object it operates against, and if this evaluates
+to `false`, the object is excluded from the index / results. You can implement a `getShowInSearch` method on your
+DataObject to control the way this is computed. This check happens in two places:
+
+a) When attempting to add the object to the search index (or update it)
+b) Before returning results from the search index. Note: this only applies to Solr 4 implementations.
+
+The second check is an additional layer to ensure that a result is excluded if the evaluated response changes between
+index and query time. For example, a getShowInSearch() implementation that filters out objects after a certain date
+might return `true` when the object is added to the index, but `false` when a user later performs a search.
+
+This filtering is applied to all Page (SiteTree) and File records since they have a ShowInSearch database column.
+This will also be applied to any DataObjects that have a ShowInSearch database column or a getShowInSearch() function.
+
+This is a compulsory check and there is no opt-out available.
+
+Note: If you implement a custom getShowInSearch() method on a Page, the database column 'ShowInSearch' will not be used
+and the 'Show In Search?' settings in the CMS admin found under Page > Settings will no longer work. Either incorporate
+the ShowInSearch column in your getShowInSearch() logic, or remove the field from the CMS to minimise confusion.
 
 ## Solr dev tasks
 
--- a/src/Search/Processors/SearchUpdateProcessor.php
+++ b/src/Search/Processors/SearchUpdateProcessor.php
@@ -2,6 +2,7 @@
 
 namespace SilverStripe\FullTextSearch\Search\Processors;
 
+use SilverStripe\FullTextSearch\Search\Services\IndexableService;
 use SilverStripe\ORM\DataObject;
 use SilverStripe\FullTextSearch\Search\Variants\SearchVariant;
 use SilverStripe\FullTextSearch\Search\FullTextSearch;
@@ -31,7 +32,7 @@ abstract class SearchUpdateProcessor
      * @var array
      */
     protected $dirty;
-    
+
     public function __construct()
     {
         $this->dirty = array();
@@ -59,7 +60,7 @@ abstract class SearchUpdateProcessor
 
         $this->dirty[$base] = $forclass;
     }
-    
+
     /**
      * Generates the list of indexes to process for the dirty items
      *
@@ -71,6 +72,7 @@ abstract class SearchUpdateProcessor
         $dirtyIndexes = array();
         $dirty = $this->getSource();
         $indexes = FullTextSearch::get_indexes();
+        $indexableService = IndexableService::singleton();
         foreach ($dirty as $base => $statefulids) {
             if (!$statefulids) {
                 continue;
@@ -87,7 +89,12 @@ abstract class SearchUpdateProcessor
                 foreach ($objs as $obj) {
                     foreach ($ids[$obj->ID] as $index) {
                         if (!$indexes[$index]->variantStateExcluded($state)) {
-                            $indexes[$index]->add($obj);
+                            // Remove any existing records from index if ShowInSearch is changed to false
+                            if (!$indexableService->isIndexable($obj)) {
+                                $indexes[$index]->delete($base, $obj->ID, $state);
+                            } else {
+                                $indexes[$index]->add($obj);
+                            }
                             $dirtyIndexes[$index] = $indexes[$index];
                         }
                     }
@@ -105,11 +112,11 @@ abstract class SearchUpdateProcessor
                 }
             }
         }
-        
+
         SearchVariant::activate_state($originalState);
         return $dirtyIndexes;
     }
-    
+
     /**
      * Commits the specified index to the Solr service
      *
@@ -120,7 +127,7 @@ abstract class SearchUpdateProcessor
     {
         return $index->commit() !== false;
     }
-    
+
     /**
      * Gets the record data source to process
      *
--- a/src/Search/Services/IndexableService.php
+++ b/src/Search/Services/IndexableService.php
@@ -1,0 +1,58 @@
+<?php
+
+namespace SilverStripe\FullTextSearch\Search\Services;
+
+use SilverStripe\Core\Extensible;
+use SilverStripe\Core\Injector\Injectable;
+use SilverStripe\ORM\DataObject;
+use SilverStripe\ORM\Tests\MySQLDatabaseTest\Data;
+
+/**
+ * Checks if a DataObject is publically viewable thus able to be added or retrieved from a publically searchable index
+ * Caching results because these checks may be done multiple times as there a few different code paths that search
+ * results might follow in real-world search implementations
+ */
+class IndexableService
+{
+
+    use Injectable;
+    use Extensible;
+
+    protected $cache = [];
+
+    public function clearCache(): void
+    {
+        $this->cache = [];
+    }
+
+    public function isIndexable(DataObject $obj): bool
+    {
+        // check if is a valid DataObject that has been persisted to the database
+        if (is_null($obj) || !$obj->ID) {
+            return false;
+        }
+
+        $key = $this->getCacheKey($obj);
+        if (isset($this->cache[$key])) {
+            return $this->cache[$key];
+        }
+
+        $value = true;
+
+        // This will also call $obj->getShowInSearch() if it exists
+        if (isset($obj->ShowInSearch) && !$obj->ShowInSearch) {
+            $value = false;
+        }
+
+        $this->extend('updateIsIndexable', $obj, $value);
+        $this->cache[$key] = $value;
+        return $value;
+    }
+
+    protected function getCacheKey(DataObject $obj): string
+    {
+        $key = $obj->ClassName . '_' . $obj->ID;
+        $this->extend('updateCacheKey', $obj, $key);
+        return $key;
+    }
+}
--- a/src/Solr/Reindex/Handlers/SolrReindexBase.php
+++ b/src/Solr/Reindex/Handlers/SolrReindexBase.php
@@ -4,6 +4,7 @@ namespace SilverStripe\FullTextSearch\Solr\Reindex\Handlers;
 
 use Psr\Log\LoggerInterface;
 use SilverStripe\Core\Environment;
+use SilverStripe\FullTextSearch\Search\Services\IndexableService;
 use SilverStripe\FullTextSearch\Solr\Solr;
 use SilverStripe\FullTextSearch\Solr\SolrIndex;
 use SilverStripe\FullTextSearch\Search\Variants\SearchVariant;
@@ -220,6 +221,7 @@ abstract class SolrReindexBase implements SolrReindexHandler
     {
         // Generate filtered list of local records
         $baseClass = DataObject::getSchema()->baseDataClass($class);
+        /** @var DataList $items */
         $items = DataList::create($class)
             ->where(sprintf(
                 '"%s"."ID" %% \'%d\' = \'%d\'',
@@ -236,6 +238,20 @@ abstract class SolrReindexBase implements SolrReindexHandler
             $items = $items->filter('ClassName', $class);
         }
 
+        $indexableService = IndexableService::singleton();
+
+        // ShowInSearch filter
+        // we cannot use $items->remove($item), as that deletes the record from the database
+        $idsToRemove = [];
+        foreach ($items as $item) {
+            if (!$indexableService->isIndexable($item)) {
+                $idsToRemove[] = $item->ID;
+            }
+        }
+        if (!empty($idsToRemove)) {
+            sort($idsToRemove);
+            $items = $items->exclude(['ID' => $idsToRemove]);
+        }
         return $items;
     }
 
--- a/src/Solr/SolrIndex.php
+++ b/src/Solr/SolrIndex.php
@@ -9,6 +9,7 @@ use SilverStripe\FullTextSearch\Search\Indexes\SearchIndex;
 use SilverStripe\FullTextSearch\Search\Queries\SearchQuery;
 use SilverStripe\FullTextSearch\Search\Queries\SearchQuery_Range;
 use SilverStripe\FullTextSearch\Search\SearchIntrospection;
+use SilverStripe\FullTextSearch\Search\Services\IndexableService;
 use SilverStripe\FullTextSearch\Search\Variants\SearchVariant;
 use SilverStripe\FullTextSearch\Search\Variants\SearchVariant_Caller;
 use SilverStripe\FullTextSearch\Solr\Services\SolrService;
@@ -777,11 +778,18 @@ abstract class SolrIndex extends SearchIndex
             \Apache_Solr_Service::METHOD_POST
         );
 
+        $indexableService = IndexableService::singleton();
+
         $results = new ArrayList();
         if ($res->getHttpStatus() >= 200 && $res->getHttpStatus() < 300) {
             foreach ($res->response->docs as $doc) {
                 $result = DataObject::get_by_id($doc->ClassName, $doc->ID);
                 if ($result) {
+                    // Filter out any results previously added to the solr index where ShowInSearch == false
+                    if (!$indexableService->isIndexable($result)) {
+                        continue;
+                    }
+
                     $results->push($result);
 
                     // Add highlighting (optional)
--- a/tests/SolrIndexTest.php
+++ b/tests/SolrIndexTest.php
@@ -2,15 +2,24 @@
 
 namespace SilverStripe\FullTextSearch\Tests;
 
+use Apache_Solr_Document;
+use Page;
+use SebastianBergmann\Version;
+use SilverStripe\Assets\File;
+use SilverStripe\CMS\Model\SiteTree;
 use SilverStripe\Core\Config\Config;
 use SilverStripe\Core\Environment;
 use SilverStripe\Core\Injector\Injector;
 use SilverStripe\Core\Kernel;
 use SilverStripe\Dev\SapphireTest;
+use SilverStripe\FullTextSearch\Search\FullTextSearch;
 use SilverStripe\FullTextSearch\Search\Queries\SearchQuery;
+use SilverStripe\FullTextSearch\Search\Services\IndexableService;
+use SilverStripe\FullTextSearch\Search\Updaters\SearchUpdater;
 use SilverStripe\FullTextSearch\Search\Variants\SearchVariantSubsites;
 use SilverStripe\FullTextSearch\Solr\SolrIndex;
 use SilverStripe\FullTextSearch\Solr\Services\Solr3Service;
+use SilverStripe\FullTextSearch\Solr\Services\Solr4Service;
 use SilverStripe\FullTextSearch\Tests\SearchUpdaterTest\SearchUpdaterTest_Container;
 use SilverStripe\FullTextSearch\Tests\SearchUpdaterTest\SearchUpdaterTest_HasOne;
 use SilverStripe\FullTextSearch\Tests\SearchUpdaterTest\SearchUpdaterTest_HasMany;
@@ -21,10 +30,25 @@ use SilverStripe\FullTextSearch\Tests\SolrIndexTest\SolrIndexTest_AmbiguousRelat
 use SilverStripe\FullTextSearch\Tests\SolrIndexTest\SolrIndexTest_BoostedIndex;
 use SilverStripe\FullTextSearch\Tests\SolrIndexTest\SolrIndexTest_FakeIndex;
 use SilverStripe\FullTextSearch\Tests\SolrIndexTest\SolrIndexTest_FakeIndex2;
+use SilverStripe\FullTextSearch\Tests\SolrIndexTest\SolrIndexTest_ShowInSearchIndex;
+use SilverStripe\FullTextSearch\Tests\SolrIndexTest\SolrIndexTest_MyPage;
+use SilverStripe\FullTextSearch\Tests\SolrIndexTest\SolrIndexTest_MyDataObjectOne;
+use SilverStripe\FullTextSearch\Tests\SolrIndexTest\SolrIndexTest_MyDataObjectTwo;
+use SilverStripe\ORM\DataObject;
 use SilverStripe\Subsites\Model\Subsite;
+use SilverStripe\Versioned\Versioned;
 
 class SolrIndexTest extends SapphireTest
 {
+
+    protected $usesDatabase = true;
+
+    protected static $extra_dataobjects = [
+        SolrIndexTest_MyPage::class,
+        SolrIndexTest_MyDataObjectOne::class,
+        SolrIndexTest_MyDataObjectTwo::class,
+    ];
+
     public function testFieldDataHasOne()
     {
         $index = new SolrIndexTest_FakeIndex();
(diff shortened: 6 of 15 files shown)


src/Search/Processors/SearchUpdateProcessor.php
```diff
@@ -83,12 +83,19 @@ protected function prepareIndexes()
...
                 // Ensure that indexes for all new / updated objects are included
+                $indexesThatDeleted = [];
                 $objs = DataObject::get($base)->byIDs(array_keys($ids));
                 foreach ($objs as $obj) {
                     foreach ($ids[$obj->ID] as $index) {
                         if (!$indexes[$index]->variantStateExcluded($state)) {
-                            $indexes[$index]->add($obj);
-                            $dirtyIndexes[$index] = $indexes[$index];
+                            // remove any existing records from search index if ShowInSearch is false
+                            if (isset($obj->ShowInSearch) && !$obj->ShowInSearch && $obj->ID) {
```
chillu: `isset()` means it won't fall back to `$obj->getShowInSearch()` right? Also, the check for `$obj->ID` is a bit confusing here - do we support indexing recirds without an ID (before they're saved)? I wouldn't know how we'd reference them in the index then. Seems superfluous.

src/Search/Processors/SearchUpdateProcessor.php
```diff
@@ -104,6 +111,10 @@ protected function prepareIndexes()
                     }
                 }
             }
+
+            foreach (array_keys($indexesThatDeleted) as $index) {
+                $indexes[$index]->commit();
```
chillu: Should use the highlevel API provided here - `commitIndex()`.

So in case an index is both "dirty" and had deletions (which is quite likely), commit is now called twice - once in `prepareIndexes()`, and once in the callee of that (`process()`). I assume since there's nothing to commit the second time around, that's a pretty cheap operation. But it feels hacky. For starters, if the new commits fail, the parent method won't return false. `prepareIndexes()` is a protected method, so technically it's an API change to return *both* index sets, but I think it's consistent with the meaning of the method, so would advocate to just use the existing `commit()` invocation.

src/Solr/Reindex/Handlers/SolrReindexBase.php
```diff
@@ -236,6 +237,17 @@ protected function getRecordsInGroup(SolrIndex $indexInstance, $class, $groups,
             $items = $items->filter('ClassName', $class);
         }
 
+        // ShowInSearch filter
+        // we cannot use $items->remove($item), as that deletes the record from the database
+        $ids = $items->sort('ID')->column('ID');
+        $ids = array_fill_keys($ids, true);
+        foreach ($items as $item) {
+            if (isset($item->ShowInSearch) && !$item->ShowInSearch) {
```
chillu: Same concern here, that likely ignores any overloaded getters?

src/Solr/SolrIndex.php
```diff
@@ -779,6 +779,12 @@ public function search(SearchQuery $query, $offset = -1, $limit = -1, $params =
             foreach ($res->response->docs as $doc) {
                 $result = DataObject::get_by_id($doc->ClassName, $doc->ID);
                 if ($result) {
+                    // Filter out any results previously added to the solr index that have ShowInSearch == 0
+                    // This may also include any results where getShowInSearch() == false
+                    if (isset($result->ShowInSearch) && !$result->ShowInSearch) {
```
chillu: This should note that it's a second level check, and we already prevent those records from being indexed in most cases. It also means the `$offset` and `$limit` args of this method might no longer be applicable, in the worst case you've asked for 20 results and you get zero because they've all been filtered out.

I think this is too low level, and should happen on the same level where we already do the canView() filtering (and devs already need to expect the issue raised above) - in `SearchForm->getResults()` and `CwpSearchEngine`. If you subclass that and replace the implementation you're on your own.

emteknetnz: >> It also means the $offset and $limit args of this method might no longer be applicable
I don't think there's anyway around this because we need to support getShowInSearch(), so we cannot rely solely on solr filtering which would allow us to preserve $offet and $limt.  This is really no different from how canView() has been all this time.

>> I think this is too low level
How come?  It seems like the only downside here is that we now make it really difficult for people to NOT run the ShowInSearch check

src/Search/Processors/SearchUpdateProcessor.php
emteknetnz: - Should use the highlevel API provided here - commitIndex()
- so would advocate to just use the existing commit() invocation.

I'm a little unsure what you're asking me to do here, could you please clarify

src/Search/Processors/SearchUpdateProcessor.php
emteknetnz: have removed the $obj->ID check

isset() does actually work for getShowInSearch() - see https://github.com/silverstripe/silverstripe-fulltextsearch/issues/257#issuecomment-601488939

src/Solr/Reindex/Handlers/SolrReindexBase.php
emteknetnz: overloaded getters still work, see https://github.com/silverstripe/silverstripe-fulltextsearch/issues/257#issuecomment-601488939

src/Search/Processors/SearchUpdateProcessor.php
chillu: Right, we've got `ViewableData->__isset()`.

src/Search/Processors/SearchUpdateProcessor.php
chillu: Hah sorry, I can see how that's confusing. The important bit here is that there's only one commit per index for *both* indexes with dirty states and deletes. `commitIndex()` already gets called in `process()`. You shouldn't need to call `commit()` separately yourself, you should adjust the `prepareIndexes()` method return value instead to incorporate the additional indexes.

src/Solr/Reindex/Handlers/SolrReindexBase.php
```diff
@@ -236,6 +237,17 @@ protected function getRecordsInGroup(SolrIndex $indexInstance, $class, $groups,
             $items = $items->filter('ClassName', $class);
         }
 
+        // ShowInSearch filter
+        // we cannot use $items->remove($item), as that deletes the record from the database
+        $ids = $items->sort('ID')->column('ID');
```
emteknetnz: $ids = $items->map('ID', 'ID')->toArray();

src/Solr/SolrIndex.php
chillu: We've talked through this on a call, results in https://github.com/silverstripe/silverstripe-fulltextsearch/issues/257#issuecomment-605717864

src/Search/Processors/SearchUpdateProcessor.php
```diff
@@ -83,12 +83,19 @@ protected function prepareIndexes()
                 SearchVariant::activate_state($state);
 
                 // Ensure that indexes for all new / updated objects are included
+                $indexesWithDeletes = [];
                 $objs = DataObject::get($base)->byIDs(array_keys($ids));
```
emteknetnz: Unfortunately $base is going to be SiteTree, this means that any subclasses (e.g. MyPage) will be retrieved as SiteTree instead, so MyPage->getShowInSearch() won't be called :-(

This happens even if I explicitly add this to the index:
$this->addClass(MyPage::class);

chillu: `addDirtyIDs ()` has this: `$base = DataObject::getSchema()->baseDataClass($class);`, which is then used in `prepareIndexes()`. So at some point, the specific class was available in this implementation, it's just a matter of surfacing it in the intermediary array?

emteknetnz: yup ignore me, I'm not sure how I came to this conclusion.  I've written a unit test that explicitly tests for getShowInSearch() on a couple of classes and that's passing

src/Search/Processors/SearchUpdateProcessor.php
emteknetnz: ok, have removed the double commit. have simply ensured that `$dirtyIndexes[$index] = $indexes[$index];` is run on both add + delete

src/Solr/SolrIndex.php
Cheddam: I agree with @chillu about the placement of this code - better to have a consistent set of conditional logic incorporating both ShowInSearch and canView throughout the module. If you're using high-level APIs, you're covered, if you're using low-level APIs, the documentation should outline the risks and you have a clearer example to copy if the checks are neighboured.

emteknetnz: I think am in agreement here.  Reasoning is slightly different though.  I know there are plenty of custom search implementations in the wild that copy-pasted the SolrIndex->search() function into their own MyCustomSolrIndex because they want to tweak something e.g. highlighting.  Because of this, they don't call parent::search().  This means that any change we make here won't make it into their implementation.

src/Solr/Reindex/Handlers/SolrReindexBase.php
```diff
@@ -236,6 +237,20 @@ protected function getRecordsInGroup(SolrIndex $indexInstance, $class, $groups,
...
+        foreach ($items as $item) {
+            if (isset($item->ShowInSearch) && !$item->ShowInSearch) {
+                unset($ids[$item->ID]);
+            }
+        }
+        $items = DataList::create($class);
+        // test item count first to prevent exception: Cannot filter "SiteTree"."ID" against an empty set
+        if (count($ids) > 0) {
+            return $items->filter(['ID' => array_keys($ids)]);
+        }
```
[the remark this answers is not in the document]
emteknetnz: good catch, it will return everything by default

src/Solr/Reindex/Handlers/SolrReindexBase.php
```diff
@@ -239,19 +239,15 @@ protected function getRecordsInGroup(SolrIndex $indexInstance, $class, $groups,
...
             }
         }
-        $items = DataList::create($class);
-        // test item count first to prevent exception: Cannot filter "SiteTree"."ID" against an empty set
-        if (count($ids) > 0) {
-            return $items->filter(['ID' => array_keys($ids)]);
-        }
-        return $items;
+        // If no $ids, then filter against 0 to prevent exception: Cannot filter "SiteTree"."ID" against an empty set
+        $value = count($ids) > 0 ? array_keys($ids) : 0;
```
dhensby, line 249:
```suggestion
        $value = !empty ($ids) ? array_keys($ids) : 0;
```

src/Solr/SolrIndex.php
chillu: > they don't call parent::search(). This means that any change we make here won't make it into their implementation.

That's a risk I think we have to take, in alignment with the priorities I've outlined. We should avoid APIs that force devs to copy paste, and rely on composability instead (e.g. not forcing devs to mix in presentational highlighting concerns with core permission checks). Let's capture this in well written upgrade guides, outlining what happen before, what we've changed, and in which situations you could be vulnerable even after an upgrade.

docs/en/03_configuration.md
```diff
@@ -172,6 +172,24 @@ you've just created, this will add `SearchableDataObject` and the text fields it
...
 
+### ShowInSearch and getShowInSearch() filtering
+
+A call to $dataObject->ShowInSearch, which can be overloaded with $dataObject->getShowInSearch() is made
+in the fulltextsearch module so that ShowInSearch == false results are automatically filtered in two places:
+a) Before adding data to the search index
+b) Before returning results from the search index. Note: this only applies to Solr 4 implementations.
+
+The second check is a second layer for special scenarios such as a getShowInSearch() implementation that filters out
+results after a hardcoded date i.e. returns true before adding to the index, returns false some time later.
```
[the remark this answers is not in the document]
emteknetnz: Thanks for feedback on docs, have implemented all your recommended changes

src/Solr/Reindex/Handlers/SolrReindexBase.php
```diff
@@ -236,6 +237,20 @@ protected function getRecordsInGroup(SolrIndex $indexInstance, $class, $groups,
             $items = $items->filter('ClassName', $class);
         }
 
+        // ShowInSearch filter
+        // we cannot use $items->remove($item), as that deletes the record from the database
+        $ids = $items->sort('ID')->column('ID');
+        $ids = array_fill_keys($ids, true);
```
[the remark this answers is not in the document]
emteknetnz: Have implemented

src/Solr/Reindex/Handlers/SolrReindexBase.php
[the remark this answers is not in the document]
emteknetnz: have fixed

src/Search/Extensions/IndexableExtension.php
```diff
@@ -0,0 +1,46 @@
...
+    protected $hasCheckedIsIndexable = false;
+
+    protected $canBeIndexed = false;
+
+    /**
+     * @return bool
+     */
+    public function isIndexable()
+    {
+        if ($this->owner->hasCheckedIsIndexable) {
```
chillu, line 22: You could also do this as a tri-state, and check for `canBeIndexed === NULL`. The advantage of that is that it's easier to perform test setups mainly. And you don't get into awkward naming disparities like "can be indexed" vs "has checked is indexable" which are correlated, but not in naming.

emteknetnz: Changed to a tri-state.  I used const instead of nullable boolean for a bit more clarify and less likely to make mistakes with `if (!$this->owner->indexedState)`

src/Search/Extensions/IndexableExtension.php
```diff
@@ -0,0 +1,53 @@
...
+    const INDEXABLE_STATE_UNCHECKED = 'INDEXABLE_STATE_UNCHECKED';
+
+    const INDEXABLE_STATE_ALLOWED = 'INDEXABLE_STATE_ALLOWED';
+
+    const INDEXABLE_STATE_DISALLOWED = 'INDEXABLE_STATE_DISALLOWED';
+
+    /**
+     * @var string
+     */
+    protected $indexableState;
```
Cheddam, line 21: I'm trying to comprehend why this is a safe place to store this state - are `DataExtension` objects persisted against `DataObjects` across multiple queries / `DataLists` somehow? If not, this runs the risk of being recomputed in cases where it's called multiple times at different layers. Not sure if that statement holds up in real-world cases, just a little concerned.

src/Search/Extensions/IndexableExtension.php
```diff
@@ -0,0 +1,53 @@
...
+            return false;
+        }
+
+        // TODO: check if class is omitted from canView() check
+        // TODO: Member::actAs(null, ... to test an an anonymous user .. possibly make this a function param
+        // TODO: uncomment - breaks unit tests as is
+//        if (!$this->owner->canView()) {
+//            $this->owner->indexableState = self::INDEXABLE_STATE_DISALLOWED;
+//            return false;
+//        }
```
Cheddam, line 48: Let's drop this out before merging, I'll reintroduce it in a functioning state in my PR.

tests/IndexableExtensionTest.php
```diff
@@ -0,0 +1,28 @@
...
+        $page = SiteTree::create();
+        $page->CanViewType = 'Anyone';
+        $page->ShowInSearch = 0;
+        $this->assertFalse($page->isIndexable());
+
+        // TODO: renable once canView() check is uncommented from IndexableExtension
+//        $page = SiteTree::create();
+//        $page->CanViewType = 'LoggedInUsers';
+//        $page->ShowInSearch = 1;
+//        $this->assertFalse($page->isIndexable());
```
Cheddam, line 26: This can be removed and reintroduced in my `canView` PR.

Cheddam:
Sidenote - this needs to be rebased and retargeted to the `3` branch, as it introduces new API surface.

src/Search/Extensions/IndexableExtension.php
```diff
@@ -0,0 +1,53 @@
+<?php
+
+namespace SilverStripe\FullTextSearch\Search\Extensions;
+
+use SilverStripe\ORM\DataExtension;
+
+/**
+ * Checks if a DataObject is publically viewable thus able to be added or retrieved from a publically searchable index
+ */
+class IndexableExtension extends DataExtension
```
chillu, line 10: Making this a default extension on `DataObject` means more overhead for every method call, and framework boot times. It also makes this hard to extend by modules and custom implementations. For example, how would subsites indicate that a record shouldn't be indexed if a subsite has been "disabled" (hypothetically). I'd make this a service instead, and by extending ViewableData you can allow extension hooks on it. By checking for an `isIndexable` method on the DataObject itself, you can also allow more fine tuning of this functionality in a way that's hard through an extension. You're only using public methods on DataObjects here anyway. By making it a singleton, you can add cache state (`$indexableState`) in the service rather than on the DataObject

src/Search/Extensions/IndexableExtension.php
emteknetnz: removed

emteknetnz:
@Cheddam have rebased and retargeted to 3 (used to be 3.6) cos as you say, new API = minor release

src/Search/Extensions/IndexableExtension.php
emteknetnz: Ingo recommended changing this from a DataExtension to a singleton instead to hold all the cached results, so will do that

src/Search/Extensions/IndexableExtension.php
emteknetnz: @chillu 

service/singleton makes sense, will do that

> by extending ViewableData 
hmm, a bit strange.  reads weird and will end up with a bunch of unnecessary function.  wouldn't it be cleaner to just add the Extensible trait directly (along with the Injectable trait)?

So regarding making this extendable, is this how you see it working?

```
class IndexableService
..
use Extensible;
... 
public function isIndexable(DataObject $obj)
{
    $result = [do stuff]
    $this->extend('updateIsIndexable', $obj, $result);
}
```

Add in say subsites module:

extensions.yml
```
---
Name: subsitesfulltextsearch
Only:
  moduleexists: 'silverstripe/fulltextsearch'
---
SilverStripe\FullTextSearch\Search\Services\IndexableService:
  extensions:
    - SilverStripe\Subsites\Extensions\SubsitesIndexableServiceExtension
```

SubsitesIndexableServiceExtension.php
```
class SubsitesIndexableServiceExtension extends Extension
...
public function updateIsIndexable(&obj, &result)
{
    if ($subsiteDisabled) {
        $result = false;
```

Let me know if my thinking is correct or needs adjusting

Cheddam: That all makes sense to me. Agree that using `Extensible` / `Injectable` / `Configurable` traits should be sufficient.

src/Solr/Reindex/Handlers/SolrReindexBase.php
```diff
@@ -99,6 +100,8 @@ protected function processVariant(
         $batchSize,
         $taskName
     ) {
+        // Get current state
+        $originalState = SearchVariant::current_state();
```
emteknetnz, line 104: These `$originalState` commits are from a different PR and showed up after I rebased on to `3`, was originally targetted `3.6` I think what happened is there was a missing merge-up

src/Search/Extensions/IndexableExtension.php
emteknetnz: Have refactored this from a DataExtension to a Singleton/Service

chillu: Sorry I mean `Extensible`, this comes from an old school time when this was all lumped into `ViewableData` back in 3.x :D `extend()` also by default allows for NULL returns, so you can have extensions "abstain" from a decision which can be important. Same logic as in canView etc

src/Search/Services/IndexableService.php
```diff
@@ -0,0 +1,49 @@
...
+    use Extensible;
+
+    protected $cache = [];
+
+    public function clearCache(): void
+    {
+        $this->cache = [];
+    }
+
+    public function isIndexable(DataObject $obj): bool
```
chillu, line 26: So this wouldn't work if e.g. one language variation of an object is indexable but the other one isn't. Funnily enough this PR from Terraformers might come in handy for this: https://github.com/silverstripe/silverstripe-framework/pull/9400. For now, it's good enough

src/Solr/Reindex/Handlers/SolrReindexBase.php
```diff
@@ -224,7 +238,19 @@ protected function getRecordsInGroup(SolrIndex $indexInstance, $class, $groups,
...
+        // we cannot use $items->remove($item), as that deletes the record from the database
+        $ids = $items->map('ID', 'ID')->toArray();
+        foreach ($items as $item) {
+            if (!$indexableService->isIndexable($item)) {
+                unset($ids[$item->ID]);
+            }
+        }
+        // If no $ids, then filter against 0 to prevent exception: Cannot filter "SiteTree"."ID" against an empty set
+        $value = !empty($ids) ? array_keys($ids) : 0;
+        return DataList::create($class)->filter(['ID' => $value]);
```
chillu, line 253: This feels like a bit of a foot gun: You're re-creating the $items list above as a separate DataList, rather than filtering out from the existing list. I'd reframe this to `$items->exclude('ID', $removedIds)`

tests/SolrIndexTest.php
```diff
@@ -376,6 +398,94 @@ public function testGetIndexNameWithPrefixAndSuffixFromEnvironment()
...
+        // will get added
+        // this class has a getShowInSearch() == true, which will override $mypage->ShowInSearch = false
+        $mypage = new SolrIndexTest_MyPage();
+        $mypage->Title = 'Test MyPage false';
+        $mypage->ShowInSearch = false;
+        $mypage->write();
+
+        // 1x Page + 1x File + 1x SolrIndexTest_MyPage,
+        // 1x SolrIndexTest_MyDataObjectOne + 1x SolrIndexTest_MyDataObjectTwo
+        $serviceMock->expects($this->exactly(5))->method('addDocument');
```
chillu, line 473: That's a bit of a weak test, it would also pass if the first object is added five times, and the other ones ignored. I would break down test methods to a more granular level, as a general comment. These long test setups are really hard to read.

src/Search/Services/IndexableService.php
emteknetnz: you could modify this via `$this->extend('updateIsIndexable', $obj, $value);` ?

e.g.
```
public function updateIsIndexable(&$obj, &$value) {
   if ($obj->IsEnglish) {
    $value = true;
   } elseif ($obj->IsFrench) {
    $value = false;
  } 
}
```
?

emteknetnz: Though, I guess if you have $obj->EnglishContent and $obj->FrenchContent, then it won't be so good

Would it be better if we cache by default and have a Configurable opt-out?  Would mean that you'd need to do (apologies if I have syntax wrong) $this->config()->get('do_cache_is_indexable') on every check, which would lead to some level of performance degredation

src/Solr/Reindex/Handlers/SolrReindexBase.php
emteknetnz: Have changed it to use exclude()

tests/SolrIndexTest.php
emteknetnz: Have implemented testing the parameters passed to the $mock->addDocument() so we no longer risk passing it when it's actually failing

src/Solr/Reindex/Handlers/SolrReindexBase.php
Cheddam: Hmm, we'll want to tidy this up before merging. You get the fun task of counting how many commits you've made and running `git rebase --onto=3 HEAD~x` 😅

src/Search/Services/IndexableService.php
Cheddam: Is this worth addressing upfront, rather than waiting for it to impact a project? Do we have a gauge on how Fluent is currently used with fulltextsearch – do projects use separate indexes for each locale?

src/Solr/Reindex/Handlers/SolrReindexBase.php
emteknetnz: ok sorted, thanks for the merge-up @dhensby

src/Search/Services/IndexableService.php
chillu: There's not even a search variant class for Fluent (like for [subsites](https://github.com/silverstripe/silverstripe-fulltextsearch/blob/3/src/Search/Variants/SearchVariantSubsites.php)).

I assume that when you call reindex on a site with fluent, you already have to perform some customisations to that indexing logic.  We could just reset the cache somewhere around here? https://github.com/silverstripe/silverstripe-fulltextsearch/blob/277a307a1b225587e93f69d86c3f5f2817a97f3d/src/Solr/Reindex/Handlers/SolrReindexBase.php#L55

chillu: This might also be an argument *against* employing a cache here. Can't we just ensure that isIndexable() is only called once per record/variant on an indexing request? ShowInSearch is cheap, canView is expensive. So worth caching, but only if it's called multiple times.

emteknetnz: >>  Can't we just ensure that isIndexable() is only called once per record/variant on an indexing request

- Outlet filtering - we cannot ensure isIndexable() is only called once
- Inlet filtering we can be fairly confident that it's only being called once

Outlet filtering is messy because that's where we see the most variation with custom implementations.  Because of this we've needed to put isIndexable() is multiple places because there is central place (that's still sane) that all requests in all custom implementations from the solr server go through

e.g.
- Scenario 1. Uses SearchForm.php, CustomSolrIndex->search() calls parent::search() - isIndexable() called twice
- Scenario 2. Does not use SearchForm.php, instead calls CustomSolrIndex->search() from in a page template (probably did not follow docs :-)) - isIndexable() called once
- Scenario 3. Uses SearchForm.php, CustomSolrIndex->search() is copy pasted from SolrIndex, does not call parent::search() - isIndexable() called once

Inlet filtering on the other hand is much easier since it's either SearchUpdateProcessor (cms->save() or $dataObject->write() or SolrReindexBase (dev/tasks/Solr_Reindex), bother of which are very unlikely to be customised

A good argument to remove caching from the service would be that it's only there for the sake of outlet filtering, there it's likely to be paginated to say 10 results, so you're looking at 20 vs 10 canView() calls, so the performance hit may not be too bad

I can see a few options here:
a) Remove caching from the service and accept the outlet filtering performance hit (most secure and easiest)
b) Add the Configurable trait and make caching opt-out so that modules such as fluent can disable it (though those modules will themselves need to be tested and updated, creates a lot of work)
c) Test for the existence modules such as fluent from a hardcoded list within this module and if class_exists(Fluent), then we disable caching
d) Keep things as they are now, accept some level of risk that isIndexable() will return the wrong result in some scenarios we haven't yet though of.

I don't really see removing isIndexable() from multiple locations as a valid option, just because custom search implementations didn't quite do things by the book doesn't seem like a good enough justification to shift responsibility to them for things going wrong.

I don't like b) or c).  Too complicated.

a) is attractive because it'll just work, will result in front end search results being slower though
d) is probably fine.  Inlet filtering will largely prevent anything that shouldn't be there getting indexed in the first place.  Outlet filtering is the second layer of defense.

Let me know what option you think is best.

chillu: Hey Steve, thanks for working through this, great TED talk would read again ;)
How about a separate `getCacheKey($obj)` which can then be augmented by `updateCacheKey($obj)`? The getter then simply concats any additional cache keys. In some cases cache keys in extensions will be dependant on wider context (versioned stage, active fluent locale), but any extension can read this context. We wouldn't have to implement it for Fluent etc, just ensure that it *can* be done. This is somewhat of a temporary API anyway and should be superseded with `DataObject->getUniqueKey()` once that's merged (https://github.com/silverstripe/silverstripe-framework/pull/9400). But we need to ensure backwards compatibility here.

emteknetnz: Sounds good.  Have implemented.  I'll mark this one as resolved.

PR #276, opened 2020-03-29
WIP: NEW Respect canView permissions during index operations

Cheddam:
The module now calls `canView()` on documents to determine whether they should exist in the index at all, no longer solely relying on this check occurring post-search. This decreases the likelihood of data leakage, and should result in more accurate search result pagination.

This new logic attempts to make use of `InheritedPermissions::canViewMultiple()` when the DataObject exposes it via `getPermissionChecker()`. This results in less accurate output, as it's based solely on permissions stored in the database, but should be sufficient for most models that are using it.

The new check can be disabled on a per-class basis (will also disable descendants), for situations where it's too expensive or known not to be necessary.

This PR also bumps the minimum PHP requirement for this module to 7.1, to enable use of modern language features.

#### Performance testing

Initial benchmarks (native PHP 7.1 / MySQL 8 on a mid-range 15" 2018 MacBook Pro, running against data from the default frameworktest population script, which generates ~2000 Page objects):

```
canView Disabled
12.10s user 5.63s system 79% cpu 22.353 total
12.15s user 5.64s system 77% cpu 22.970 total
12.21s user 5.66s system 73% cpu 24.229 total

canView Enabled (with canViewMultiple cache)
12.19s user 5.48s system 80% cpu 22.063 total
12.34s user 5.66s system 80% cpu 22.382 total
12.51s user 5.62s system 71% cpu 25.334 total

canView Enabled (without canViewMultiple cache)
21.97s user 6.22s system 79% cpu 35.486 total
22.06s user 6.49s system 79% cpu 35.965 total
22.74s user 7.01s system 77% cpu 38.396 total
```

#### TODO

- [ ] Add `$preindex` condition to `IndexableService::isIndexable()` method, and have the `FullTextSearch::isCanViewCheckDisabledForClass()` check apply only when this is `true`
- [ ] Consider additional `areIndexable(DataList $objs)` method to shift the `InheritedPermissions` cache warming into `IndexableService`
- [ ] Documentation
- [ ] Unit tests
- [ ] Ensure benchmarks still hold up, as implementation has changed since original tests

### DIFF
--- a/src/Search/FullTextSearch.php
+++ b/src/Search/FullTextSearch.php
@@ -15,7 +15,7 @@ class FullTextSearch
 {
     protected static $all_indexes = null;
 
-    protected static $indexes_by_subclass = array();
+    protected static $indexes_by_subclass = [];
 
     /**
      * Optional list of index names to limit to. If left empty, all subclasses of SearchIndex
@@ -24,7 +24,19 @@ class FullTextSearch
      * @var array
      * @config
      */
-    private static $indexes = array();
+    private static $indexes = [];
+
+    /**
+     * During (re)index of items, the canView method for the item is called against an anonymous user,
+     * in order to determine whether the item should be included in the index. This is a preventative
+     * security measure, and has a performance cost. If you are confident that a given DataObject should
+     * be visible to everyone, or you have other measures in place to secure the contents of the index,
+     * you can disable this check for that specific class and its descendants.
+     *
+     * @var array Should contain FQCNs of classes the check is to be disabled on (e.g. SiteTree::class)
+     * @config
+     */
+    private static $disable_preindex_canview_check = [];
 
     /**
      * Get all the instantiable search indexes (so all the user created indexes, but not the connector or library level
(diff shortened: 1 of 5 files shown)


src/Search/Processors/SearchUpdateProcessor.php
```diff
@@ -84,13 +94,35 @@ protected function prepareIndexes()
 
                 // Ensure that indexes for all new / updated objects are included
                 $objs = DataObject::get($base)->byIDs(array_keys($ids));
+                $objectPermissions = [];
+
+                // Attempt to use optimised permission checker if present.
+                // NOTE: This reduces the scope of the check to DB-based permissions (custom canView logic is ignored)
```
emteknetnz: Since this is the CMS->Save() code path, we should to just use the standard and more accurate canView() check instead of the getPermissionChecker here since performance is much less of a concern

src/Search/FullTextSearch.php
```diff
@@ -143,4 +155,24 @@ public static function force_index_list()
...
     }
+
+    /**
+     * Uses the disable_preindex_canview_check configuration to determine whether the given class should have the
+     * canView check applied.
+     *
+     * @param string $class The FQCN of the class to check
+     * @return bool
+     */
+    public static function isCanViewCheckDisabledForClass(string $class): bool
```
emteknetnz: function name reads kind of weird, `canViewCheckDisabledForClass()` might be better

src/Search/Processors/SearchUpdateProcessor.php
Cheddam: I think it makes more sense to keep the behaviour consistent across the two - there are already enough edgecases as it is. Any thoughts on this @chillu?

chillu:
Sweet, benchmarks look good. How many objecdts did you index there? Did any of them have restricted permissions, and/or inheritance? Good to see that canViewMultiple caching works, and indexing mostly "permission-less pages" is our most common use case. But that implementation would get slower e.g. if a full page tree is set to admin only, incl. children. So the data volume and structure is important to draw conclusions from the benchmark here.

Cheddam:
Superceded by #281.

PR #278, opened 2020-04-07
NEW Ensure commits are visible to seachers (fixes #274)

### LINKED ISSUE #274: Results not showing after editing pages (no soft-commit nor core reload issued)
We are seeing people having problems with their results not showing in search, after they have updated their content. The root cause of this issue is that solr index must be reloaded OR a soft-commit must happen for the results to show up. The module seems to do neither, nor it gives recommendations on how to deal with that.

Historically, soft-commit would have been triggered from queuedjobs (or core reload, can't remember). Because queuedjobs were unreliable, CWP was switched behind the scenes to hijack solrconfig.xml, and force [autoSoftCommit](https://github.com/silverstripe/silverstripe-fulltextsearch/blob/65b780f629ec31168902217d734079a14727896f/conf/solr/4/extras/solrconfig.xml#L382) to 60s, which mean the results showed reliably after saving.

I think this knowledge might have been forgotten and the module now doesn't trigger the soft commit or core reload anymore, nor even documents the necessity of committing. This means it works in CWP, but seemingly doesn't on other platforms.

I think the [default autoSoftCommit](https://github.com/silverstripe/silverstripe-fulltextsearch/blob/65b780f629ec31168902217d734079a14727896f/conf/solr/4/extras/solrconfig.xml#L382) should perhaps be changed away from -1 to something like 15000.

Maybe some documentation around handling committing could be supplied too? I can't figure out how this is supposed to work with soft-commits off by default - is there a job to run? :-) some help would be appreciated.

We can try fixing this magically on the platform side by setting the solr.autoSoftCommit.maxTime java var, but I think this will still perhaps confuse people who don't use that particular platform, so some mention in the gotchas, or docs around it would be wonderful.

chillu:
This is required in situations where autoCommit.openSearcher is set to "false",
and you don't want to rely on core restarts to make committed index updates available to new searches.
It's already configured like this in the Common Web Platform environments,
so this change will mainly affect other environments (such as Silverstripe Platform).
To opt out of this new behaviour, set up a custom solrconfig.xml.

More details in https://github.com/silverstripe/silverstripe-fulltextsearch/issues/274

### DIFF
--- a/conf/solr/4/extras/solrconfig.xml
+++ b/conf/solr/4/extras/solrconfig.xml
@@ -16,9 +16,9 @@
  limitations under the License.
 -->
 
-<!-- 
+<!--
      For more details about configurations options that may appear in
-     this file, see http://wiki.apache.org/solr/SolrConfigXml. 
+     this file, see http://wiki.apache.org/solr/SolrConfigXml.
 -->
 <config>
   <!-- In all configuration below, a prefix of "solr." for class names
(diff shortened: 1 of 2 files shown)


chillu:
Sorry for the whitespace changes, I blame PHPStorm ;) Still testing this out on SilverStripe Platform, haven't successfully indexed content through it.

Also note that I haven't changed this for Solr 3, because who the hell cares about that.

adrexia:
Is there a reason this hasn't been merged yet?

michalkleiner:
I think everyone was waiting for @chillu as he indicated in his last comment he still hadn't successfully indexed content through it, so I think everyone was looking at this as a draft PR.

Can you please confirm what's the status of this @chillu? Thanks!

emteknetnz:
I've tested this on multiple platforms and things now work as expected

emteknetnz (approved):

emteknetnz:
I've released this as [3.11.0](https://github.com/silverstripe/silverstripe-fulltextsearch/releases/tag/3.11.0)

PR #279, opened 2020-04-16
FIX #108 - Update and simplify getDirtyIDs method

### LINKED ISSUE #108: singleton($step['class'])->extend('augmentSQL', $sql); in SearchIndex::getDirtyIDs() causes throwing Fatal Errors
`singleton($step['class'])->extend('augmentSQL', $sql);` in `SearchIndex::getDirtyIDs()` function breaks some `DataExtension` objects when calling to their function augmentSQL(), eg. Translatable, or SiteTreeSubsites, since `Translatable::augumentSQL(SQLQuery &$query, DataQuery &$dataQuery = null)` is expecting the second parameter is a `DataQuery` object and directly uses it to call its function `getQueryParam()` in its implementation without checking if $dataQuery is a non-object or not

So this issue could be either fixed in Translatable class or SearchIndex class. Giving that if we fix here in SearchIndex, non of those DataExtension classes need to do anything, I hope it could fixed here.

michalkleiner:
A follow-up PR to replace #239. The tests are now green and I will test it in our project.

Thanks for the pointers @UndefinedOffset and @tractorcow!

### DIFF
--- a/src/Search/Indexes/SearchIndex.php
+++ b/src/Search/Indexes/SearchIndex.php
@@ -615,23 +615,15 @@ abstract class SearchIndex extends ViewableData
                 $ids = array($id);
 
                 foreach ($derivation['chain'] as $step) {
-                    // Use TableName for queries
-                    $tableName = DataObject::getSchema()->tableName($step['class']);
-
                     if ($step['through'] == 'has_one') {
-                        $sql = new SQLSelect('"ID"', '"' . $tableName . '"', '"' . $step['foreignkey'] . '" IN (' . implode(',', $ids) . ')');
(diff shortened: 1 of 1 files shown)


dhensby:
I'd rather get #280 merged and then merge up than have these 2 PRs that will conflict when we try to merge up

michalkleiner:
Whichever way is best. They are exactly the same, so yep, can be done that way. Potentially they wouldn't even conflict 🤷.

dhensby:
OK - I've merged up - I'm surprised that we don't see a conflict with this PR, though

michalkleiner:
I'd expect this PR now showing no changes, but perhaps I'd need to rebase and only after that it would show no changes.

Issue #252, opened 2019-04-28
Don't index records where canView=false for anonymous users


chillu:
The module denies access to a search result if you can’t view the underlying record: https://github.com/silverstripe/silverstripe-fulltextsearch/blob/f89818909a0d0c08f7d8f05217eb62e32988aa20/src/Solr/Forms/SearchForm.php#L89. It *does* however index all records, incl. ones with `canView=false`. The `canView` check is implemented in `SearchForm->getResults()`, which is a commonly subclassed method - so still a residual risk that devs will skip that check.

I think we should default to *not* indexing records where `canView()` returns `false` for anonymous users. You could overwrite this behaviour, e.g. to include `silverstripe/secureassets` in 3.x, or protected/private files in 4.x. But that would be a conscious decision, with the module staying secure by default.

We'd need to do this in a backwards compatible way (not breaking existing searches which rely on the absence of these checks). Although you could argue that those implementations rely on a bug?

When releasing this, we should strongly recommend to search users that they run a reindex to prune any records with canView=false in their current search state.

### ACs

 - [x] Records returning canView() as false for anonymous users are not indexed by default
 - [x] This behaviour applies to draft pages and other versioned records in draft stage
 - [x] This behaviour applies to draft files, as well as protected files
 - [x] A reindex prunes out any records where canView() returns false that would've been previously indexed
 - [x] When a new record is saved (unversioned) or published (versioned) in a way that changes anonymous canView() permissions, that's reflected by adding to or removing from the index
   - [x] This does not cover traversing child objects and reindexing them
 - [x] The "exclude variant" functionality is retained, but is a no-op for versioned draft stages now due to the existing canView checks
 - [x] This functionality works consistently between the core module and any CWP customisations
 - [x] Developers can opt in to their own canView handling on a per-classtree basis, with a big warning label about implications and clear documentation that minimises room for error
 - [x] There is documentation both for CWP and generic module upgrade paths
     - [x] These docs note that this implementation has edgecases and cannot be relied upon solely
 - [ ] Performance is measured against a large dataset on real-world environments (10,000 records, CWP + SC) - Steve: do we still want this? Based on https://github.com/silverstripe/silverstripe-fulltextsearch/pull/276 we can roughly expect solr_reindex to double on basic pages, though basic pages index very fast.  What really messes up reindex times is tika text-extraction on pdf's/docx's which will barely be impacted by an extra canView() check.  

### Removed ACs
[...]

PR #281, opened 2020-04-22
SearchableService canView()

emteknetnz:
Issue: https://github.com/silverstripe/silverstripe-fulltextsearch/issues/252

Adds a canView() check on everything going in to a solr index and on most implementations will also check it on the way out too

This is PR no.2 for this issue, PR no.1 (merged) added a ShowInSearch check https://github.com/silverstripe/silverstripe-fulltextsearch/pull/275

This PR in a continuation of the work previously done in this PR https://github.com/silverstripe/silverstripe-fulltextsearch/pull/276


tests/SolrReindexTest.php
```diff
@@ -394,8 +403,88 @@ public function testShowInSearch()
...
+                [$this->callback($callback)],
+                [$this->callback($callback)]
+            );
+
+        $logger = new SolrReindexTest_RecordingLogger();
+        $state = [SearchVariantVersioned::class => Versioned::DRAFT];
+        $handler = Injector::inst()->get(SolrReindexImmediateHandler::class);
+        $handler->runGroup($logger, $index, $state, SiteTree::class, 1, 0);
+        $handler->runGroup($logger, $index, $state, File::class, 1, 0);
+        $handler->runGroup($logger, $index, $state, SolrIndexTest_MyDataObjectOne::class, 1, 0);
```
Firesphere, line 487: Waste of time, tbh.

src/Search/Services/IndexableService.php
```diff
@@ -2,29 +2,53 @@
...
+
+    /**
+     * Skip the canView() check at a class level to increase performance of solr_redindex and retrieving search results.
+     * Be careful as this may lead to content showing in search results that should not be there such as non-public,
+     * cms-user-only content.  This may potentially happen via edge cases such as skipping checks where subclasses
+     * are involved.
+     *
+     * @var array namespaced classes to skip canView() check.
+     */
+    private static $classes_that_skip_canview_check = [];
```
Cheddam, line 32: This configuration was originally targeted specifically at the inlet stage, and I think it needs to change back, or at least be split in two (inlet and outlet).

There are perfectly valid use-cases for canView checks to operate only at the outlet stage - a primary example would be reintroducing the ability to search in the draft stage for logged in users. If this config is applied to both stages it will make it impossible to restore sites to the behaviour they had before upgrading to 3.7.

This will require adjustments to the `isIndexable` method to add a parameter for whether the caller is at the inlet/outlet stage, or a separate method (`isQueryable`?), to separate the behaviour.

emteknetnz: yup good point, will update

src/Search/Services/SearchableService.php
```diff
@@ -0,0 +1,162 @@
...
+
+        $value = true;
+
+        // ShowInSearch check
+        // This will also call $obj->getShowInSearch() if it exists
+        if (isset($obj->ShowInSearch) && !$obj->ShowInSearch) {
+            $value = false;
+        }
+
+        // Anonymous member canView()
```
emteknetnz, line 111: update this, only anonymous for $indexing, for search results it uses the current member

src/Search/Services/SearchableService.php
```diff
@@ -0,0 +1,162 @@
...
+        return $this->isSearchable($obj, false);
+    }
+
+    /**
+     * Checks and caches whether the given DataObject can be indexed. This is determined by two factors:
+     * - Whether the ShowInSearch property / getShowInSearch() method evaluates to true
+     * - Whether the canView method evaluates to true against an anonymous user (optional, can be disabled)
+     *
+     * @param DataObject $obj
+     * @return bool
```
emteknetnz, line 89: add bool $indexing to docblock

chillu:
> A final note: Shipping this change is going to produce manual work for every project that updates to fulltextsearch 3.7. The goal this PR strives for, of putting out any existing underground fires in sites, is likely worthy of this effort, but we need to be clear that this is an opt-out behaviour change in a minor release, and that those have burned projects before.

How do we ensure this important reminder finds its way into the highlevel CWP release announcements? So that's both the changelog and the blog post.

tests/SolrIndexTest.php
```diff
@@ -508,10 +509,105 @@ public function testShowInSearch()
...
+                File::class . $fileA->ID,
+                SolrIndexTest_MyDataObjectOne::class . $objOneA->ID
+            ];
+            return in_array($this->createSolrDocKey($doc), $validKeys);
+        };
+
+        $serviceMock
+            ->expects($this->exactly(3))
+            ->method('addDocument')
+            ->withConsecutive(
```
chillu, line 587: Have you checked that it fails if there's *more* calls? Can't tell from https://phpunit.de/manual/5.0/en/test-doubles.html#test-doubles.mock-objects

src/Search/Services/SearchableService.php
```diff
@@ -0,0 +1,162 @@
...
+     */
+    private function isSearchable(DataObject $obj, bool $indexing): bool
+    {
+        // check if is a valid DataObject that has been persisted to the database
+        if (is_null($obj) || !$obj->ID) {
+            return false;
+        }
+
+        $key = $this->getCacheKey($obj);
```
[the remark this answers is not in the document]
emteknetnz: Have added it.  Technically wasn't required because the cache is only held in memory and is non-persistent, and you're not going to do an index and search retrieving in the same request.  Still, no harm in added it and it may be helpful for anyone using the extension hooks

docs/en/03_configuration.md
```diff
@@ -114,10 +114,14 @@ If the [Queued Jobs module](https://github.com/symbiote/silverstripe-queuedjobs/
...
 
-By default, the `SearchUpdater` class indexes all available "variant states", so in the case of the `Versioned` extension, both "draft" and "live".
-For most cases, you'll want to exclude draft content from your search results.
+By default, the `SearchUpdater` class attempts to index all available "variant states", so in the case of the `Versioned` extension, it will attempt to index both "draft" and "live".
 
-You can either prevent the draft content from being indexed in the first place, by adding the following to your `SearchIndex::init()` method:
+As of version 3.7, on a default install, "draft" content will no longer be indexed by default, as it will fail an
+anonymous canView() check (since an anonymous user cannot view draft content.)
+
+For most cases, you'll want to explicitly exclude draft content from being indexed and showing in search results.  This is particularly important if you skip the canView check via SearchableService::classes_that_skip_indexing_canview_check.
```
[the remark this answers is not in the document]
emteknetnz: This section in the configuration docs is talking specifically about `$this->excludeVariantState([SearchVariantVersioned::class => Versioned::DRAFT]);`, I think as it is this section is fine

I'm thinking what might more sense to mention this in the readme, as well as the docblock for config variable to reassert that you should exclude draft content from being indexed if skip the canView check

emteknetnz:
> > A final note: Shipping this change is going to produce manual work for every project that updates to fulltextsearch 3.7. The goal this PR strives for, of putting out any existing underground fires in sites, is likely worthy of this effort, but we need to be clear that this is an opt-out behaviour change in a minor release, and that those have burned projects before.
> 
> How do we ensure this important reminder finds its way into the highlevel CWP release announcements? So that's both the changelog and the blog post.

@chillu - we've added a todo item to our future selves on the [CWP 2.6. release issue](https://github.com/silverstripeltd/product-issues/issues/168) "Work out an effective way to tell cwp users to 'run solr_reindex after composer update'"

emteknetnz:
@chillu 
> Lookin' good, left a few comments.
Cheers!

> Where did we land on with avoiding _duplicate_ canView() checks, either because they're built into a different layer in CWP, or because they're in custom code? We should advise devs on how to avoid those.
SearchableService::cache is a memory-only cache specifically for that.  The duplicate canView() don't happen when indexing `SearchableService->isIndexable($obj)`, though they can easily happen when retrieving search results  `SearchableService->isViewable($obj)` due to multiple search classes and flexibility in implementing those

> I think the lowlevel functions (SolrService->addDocument()` etc) should have a warning that they won't check if a record is indexable (not using the service)
I don't think we need to do this.  There are two indexing paths which cover all the indexing scenarios:
- cms-save/$obj->write() - goes through SearchUpdateProcessor.php
- solr_reindex - goes through - goes through SolrReindexBase.php
We do a `SearchableService->isIndexable($obj)` call in both those places

src/Search/Services/SearchableService.php
```diff
@@ -0,0 +1,162 @@
...
+        if (isset($this->cache[$key])) {
+            return $this->cache[$key];
+        }
+
+        $value = true;
+
+        // ShowInSearch check
+        // This will also call $obj->getShowInSearch() if it exists
+        if (isset($obj->ShowInSearch) && !$obj->ShowInSearch) {
+            $value = false;
```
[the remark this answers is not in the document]
emteknetnz: > ShowInSearch=false, since they also need to check canView()

That's incorrect, the canView() block below has a `if ($value)` check before doing the canView() check

ShowInSearch is inexpensive, and canView() could be potentially expensive

In terms of extensible performance, you could disable the canView check on a class via configuration and then do the canView check in your extension classes.  This will prevent an uncached duplicate canView() check so performance shouldn't be impacted.

src/Search/Services/SearchableService.php
```diff
@@ -0,0 +1,162 @@
...
+    /**
+     * Generates a unique key to cache each DataObject with. Can be extended via updateCacheKey.
+     *
+     * @param DataObject $obj
+     * @return string
+     */
+    private function getCacheKey(DataObject $obj): string
+    {
+        $key = $obj->ClassName . '_' . $obj->ID;
+        $this->extend('updateCacheKey', $obj, $key);
```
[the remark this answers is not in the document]
emteknetnz: I'm not personally familiar with the data structure that fluent uses for different language variants.  Is the english and french version of the 'Content' field effectively stored in different database columns on the same table? (so $obj->ClassName and $obj->ID are the same?)

README.md
```diff
@@ -8,9 +8,27 @@
 Adds support for fulltext search engines like Sphinx and Solr to SilverStripe CMS.
 Compatible with PHP 7.2
 
-## Maintainer Contact
+## Important note when upgrading to fulltextsearch 3.7
```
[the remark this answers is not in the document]
emteknetnz: yeah ... for 'normal' sites not doing text-extraction it's double.  Sites doing text-extraction should have a much smaller relative performance impact

README.md
```diff
@@ -8,9 +8,27 @@
...
+default against all records (DataObjects) before being added to the search index, and also before being shown in search
+results. This may mean that some records that were previously being indexed and shown in search results will no longer
+appear due to these additional checks.
+
+These additional checks have been added with data security in mind, as it's safer to assume that records failing these
+checks should not be indexed in the first place.
+
+These additional checks also mean that draft-only (unpublished) content will no longer be automatically added to the
+search index as it will now fail the canView() check for an anonymous user.  Previously draft-only content had to be
+manually excluded from being indexed via SearchIndex::excludeVariantState().
```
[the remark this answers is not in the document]
emteknetnz: >> point out how it works in CWP as a separate section

The problem there is that websites out in the wild have a fairly diverse range of search implementations.  The offical CWP docs describe the way you are "supposed to do it" and referencing that may be relevant in the majority of cases, though certainly not all cases.

>> This makes it sound like most implementations have a security flaw when they likely haven't

Fair point.  The majority of sites (though not necessary all) will have a canView() somewhere on search retrieval currently doing the security work.  I'll add a sentence mentioning this.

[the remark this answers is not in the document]
emteknetnz: Ha, oh fairly important bit I've left out.  The raw solr query contains `&fq=%2B(_versionedstage:"Live"+(*:*+-_versionedstage:[*+TO+*])` and `Live` changes to `Stage` if you suffix the url with ?stage=Stage

I'll update relevant bits in docs where we mention "draft-only"

tests/SolrIndexTest.php
emteknetnz: that's covered via `$this->exactly(3)`

changelog.md
```diff
@@ -1,90 +0,0 @@
-# Changelog
```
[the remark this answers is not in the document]
emteknetnz: We had a quick discussion on cms-squad slack channel about this.  I mentioned the module changelog is massively out of date - the last release mentioned is 2.2.0 meanwhile this module is up to 3.6.2.  We agreed that we should just prune these things since they do more harm than good being so out of date.

Our actual changelog is https://docs.silverstripe.org/en/4/changelogs/4.5.2/

If you think this is wrong and we should have up-to-date changelogs on all the modules as well then that's fine, I can undelete this changelog, though that'll be a separate issue from this pull-request

src/Search/Services/SearchableService.php
```diff
@@ -0,0 +1,162 @@
...
+     * Skip the canView() check at a class level to increase performance of search reindex.
+     * Be careful as this may lead to content showing in search results that should not be there such as non-public,
+     * cms-user-only content.  This may potentially happen via edge cases such as skipping checks where subclasses
+     * are involved.
+     *
+     * This has no effect on when search results as canView() must still be run there
+     *
+     * @var array namespaced classes to skip canView() check on search reindex
+     */
+    private static $classes_that_skip_indexing_canview_check = [];
```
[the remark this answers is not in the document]
emteknetnz: updated. yay for `replace in path` :-)

README.md
```diff
@@ -8,9 +8,27 @@
 Adds support for fulltext search engines like Sphinx and Solr to SilverStripe CMS.
 Compatible with PHP 7.2
 
-## Maintainer Contact
+## Important note when upgrading to fulltextsearch 3.7
 
-* Hamish Friedlander <hamish (at) silverstripe (dot) com>
+A canView() check against an anonymous user (i.e. someone not logged in) and a ShowInSearch check is now performed by
```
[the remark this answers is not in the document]
emteknetnz: done

src/Search/Services/SearchableService.php
```diff
@@ -0,0 +1,166 @@
...
+    /**
+     * Generates a unique key to cache each DataObject with. Can be extended via updateCacheKey.
+     *
+     * @param DataObject $obj
+     * @param bool $indexing
+     * @return string
+     */
+    private function getCacheKey(DataObject $obj, bool $indexing): string
+    {
+        $key = implode('_', [$obj->ClassName, $obj->ID, $indexing ? 'indexing' : 'viewing']);
```
emteknetnz, line 141: Should utilise https://github.com/silverstripe/silverstripe-framework/pull/9400/files which is close to being merged

Cheddam: Bear in mind that you'll also need to bump the minimum core requirement for this module to 4.6 in order to adopt that API.

> you'll never do both in the same request

Never say never 😉

emteknetnz: Fine I'll keep it :-) Will simply update to use `$obj->getUniqueKey()`

src/Search/Services/SearchableService.php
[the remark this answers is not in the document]
emteknetnz: Have switched this to use the freshly merged `$obj->getUniqueKey()`, which itself utilises `cacheKeyComponent` so I believe fluent will be covered by that?

https://github.com/silverstripe/silverstripe-framework/pull/9400/files

Have left the `updateCacheKey` extension here in case it's needed for some other unforeseen use-case, though my assumption here is that fluent/subsites/versioned/other will all be covered in `$obj->getUniqueKey()`

Is this a fair assumption to make?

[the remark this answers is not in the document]
chillu: Yes that's a fair assumption. Great to see this PR in action already, thanks

src/Search/Services/SearchableService.php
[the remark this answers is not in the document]
chillu: Right of course, I missed the `if ($value)`

README.md
```diff
@@ -8,9 +8,50 @@
...
+
+You can apply configuration to remove the new pre-index `canView()` check from your DataObjects if it is not necessary,
+or if it impedes expected functionality (e.g. for sites where users must authenticate to view any content.) This will
+also disable the check for descendants of the specified DataObjects. Ensure that your implementation of fulltextsearch
+is correctly performing a `canView()` check at query time before disabling the pre-index check, as this may result in
+leakage of private data.
+
+```
+SilverStripe\FullTextSearch\Search\Services\SearchableService:
+  classes_that_skip_indexing_canview_check:
```
chillu, line 35: Needs updated naming

README.md
```diff
@@ -8,9 +8,50 @@
...
+search results. This may mean that some records that were previously being indexed and shown in search results will
+no longer appear due to these additional checks.
+
+These additional checks have been added with data security in mind, as it's safer to assume that records failing these
+checks should not be indexed in the first place.
+
+These additional checks also mean that draft-only (unpublished) content will no longer be automatically added to the
+search index as it will now fail the `canView()` check for an anonymous user.  Previously draft-only content could be
+manually excluded from being indexed via `SearchIndex::excludeVariantState()`.  Indexed draft content was filtered from
+showing in public search results via the solr query `&fq=%2B(_versionedstage:"Live"` meaning it only requested content
```
chillu, line 24: I don't think many devs would know what to make of those raw Solr query strings - what's the intent in including it here? Just seems like noise that makes it less likely that devs will keep reading the important stuff below this point :)

README.md
```diff
@@ -8,9 +8,50 @@
...
+    - MyDataObject
+    - SiteTree # Will disable the check for all pagetypes
+```
+
+Skipping the pre-index `canView()` will mean that draft-only content will get added to the index.  This content
+will still get filtered out by the solr query `&fq=%2B(_versionedstage:"Live"`.  It's still recommended that you exclude
+draft content from being indexed if you don't need it via `SearchIndex::excludeVariantState()`.
+Read more on [excluding draft content](https://github.com/silverstripe/silverstripe-fulltextsearch/blob/3/docs/en/03_configuration.md).
+
+You can also use the updateIsSearchable extension point on SearchableService to modify the result of the method
```
chillu, line 45: Backticks

README.md
```diff
@@ -8,9 +8,50 @@
...
+after the `ShowInSearch` and `canView()` checks have run. 
+
+It is highly recommend you run a solr_reindex on your production site after upgrading from 3.6 or earlier to purge any
```
chillu, line 48: Link to documentation on how to run a full reindex?

README.md
```diff
@@ -8,9 +8,50 @@
...
+old data that should no longer be in the search index.
+
+These additional check will have an impact on the reindex performance.  It's estimated that on regular pages that
```
chillu, line 51:
```suggestion
These additional check can have an impact on the reindex performance due to additional queries for permission checks.
```

I'd leave it to the first sentence, amended above, and leave out the statement around "double time". That might be true on a stock standard recipe. I think devs can draw their own conclusions on how much impact permission checks have.

README.md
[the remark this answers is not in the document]
chillu: I still think we need more info for CWP use cases here. Point devs to the place where canView is likely called in their own logic at least (e.g. "did you subclass SearchForm? Or CwpSearchIndex?"). Give them a good chance at doing the right thing and removing the duplicate canView check from their codebase.

changelog.md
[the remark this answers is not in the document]
chillu: Hmm yeah it's a bit out of context (I'm not on that channel). So if you slip in changes like that in a PR, it's good to explain why. The best way to do that is through an isolated commit removing the changelog. I don't think now is the time to kick up a discussion about maintaining module specific changelogs, since we maintain recipes rather than modules. So tend to agree with the change itself.

README.md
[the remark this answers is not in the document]
emteknetnz: I've added two new sections:
- Details on filtering before adding content to the solr index
- Details on filtering when extracting results from the solr index

I've mentioned there's no penalty for duplicate isViewable() / canView() checks, though if they have a very custom implementation that doesn't call SearchableService::isViewable() then they should add a call in

README.md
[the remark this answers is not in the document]
emteknetnz: We now have a sentence in the readme `These additional check can have an impact on the reindex performance due to additional queries for permission checks.`

README.md
emteknetnz: Updated

README.md
emteknetnz: Have removed

README.md
emteknetnz: Have added backticks

README.md
emteknetnz: Have added link to solr reindex docs

README.md
emteknetnz: Updated as per your recommendation

emteknetnz:
> I think we need to be clear what classes are low level (some assembly required) vs. high level (secure by default). We point out e.g. on DataList where you're expected to pass raw SQL and where it's sanitised. This might be a class-level rather than method-level comment, but in general the more clarity the better when it comes to security comms towards devs.

I think we're now good here, most developers shouldn't need to change anything in their implementation after upgrading to 3.7 as it's pretty much secure by default now.  Probably the main thing is to encourage people to run SolrReindex to purge old records, though sites are no worse of then they are now if they don't.

Inlet filtering is considered the primary layer of defense, and now anything that goes through SolrReindex or DataObject->write() will now be filtered.

Outlet filtering is considered the secondary layer of defense, and I think the majority of implementations will be filtered here out of the box.  For those that are not, we have put a note in the readme "If your implementation is very custom and does not subclass nor make use of either `SolrIndex` or `SearchForm`, then it's recommended you update your implementation to call `SearchableService::isViewable()`."

emteknetnz:
@chillu do we want to release this work as a minor or a new major?  Rational for a new major would be that search results functionality may change (less results) after running `composer update`.  Rational for keeping it minor is "more security for more sites"

chillu:
> I think we're now good here

OK, I don't understand the resistance to adding a sentence worth of phpdoc to a class, but I'm not going to labour that point, your call :)

> do we want to release this work as a minor or a new major?

Ah that old chestnut. We absolutely need to get this into a minor release of the CWP recipe (via https://github.com/silverstripe/cwp-recipe-search). Otherwise the entire effort is worthless, since the point is to make current users of CWP 2.x recipe more secure by default.

On a module level, I'm in favour of a minor release. From what I can see, the "worst case scenario" for a site upgrading is that authorised users can't preview draft content in search results any more? I think that's very rare, and an acceptable change.

I guess there might be sites which are filtering out draft content already, purposely or accidentally don't check canView on search results in their custom implementation, *and* expect those results to continue being included. Which seems like an extremely odd and dangerous way of implementing this business logic (compared to respecting canView and adjusting your implementation of that accordingly).

Unless somebody can come up with a worse can of potential regressions, I'd say this is a minor for security reasons. We've worked hard on this PR to ensure it's as backwards compatible and non-intrusive as me can make it, and provide extensive and clear upgrade instructions.

brynwhyman:
@chillu I think another scenario worth considering here is existing sites who purposefully want search results to filter based on a permission check of the logged-in user, e.g. an intranet project has a 'Human Resources' parent page locked down to specific users who want to see those pages turn up in a search result.

I'd hope that we can make the possible regressions clear in the documentation and agree that this should go out as a minor. It's worth noting that this work was originally prioritised and planned to be developed as a 'type/bug' and security flaw in the module, not an enhancement.

emteknetnz:
@brynwhyman there is two options to override the default behaviour of filtering on an anonymous member canView():
- `updateIsSearchable` extension hook to add custom logic.  You can do basically whatever you like here.
- `indexing_canview_exclude_classes` configurable list of DataObjects to exclude from the canView() check on index.  canView() will still be called when viewing results.  If you add `DataObject` to this list then the search behaviour will be similar to how fulltextsearch worked previously

brynwhyman:
> I'd hope that we can make the possible regressions clear in the documentation

I just had another look at the PR... You can't miss this in the readme!

emteknetnz:
Just smoke tested this to regain context since it's been a while (not really proper testing).  It's mostly good, though seems there's one odd behaviour that we may want to correct

Good:
- Draft-only (unpublished) content is not indexed
- Pages with permissions (logged-in-users or above) is not indexed
- Subsites seems to work as expected

Weird:
- (unexpected) when you publish a Page, there are 2 records in the index, on for the draft variant, another for the live variant.  They have identical content.
- (expected) when you modify this Page and click save (but not publish), the draft-variant is deleted, so there is only a live variant, which correctly has the content on the live site

It's a bit weird, seems like when you publish a record there should be no draft-variant by default

chillu:
> when you publish a Page, there are 2 records in the index, on for the draft variant, another for the live variant. They have identical content. Adding $this->excludeVariantState([SearchVariantVersioned::class => Versioned::DRAFT]); to the custom code MySolrIndex::init() resolved

So we're effectively doing twice the indexing work and using twice the index data due to this, unless devs remember to exclude this variant which we tell them is already excluded? That's not great. Also, the index entries basically "lie" about their contents now - it claims to be draft content, but actually contains published content. I think we need to resolve this without relying on customisations in a specific project. Good catch!

brynwhyman:
> It's a bit weird, seems like when you publish a record there should be no draft-variant by default

So the unexpected behaviour is due to a core issue? 

> Adding $this->excludeVariantState([SearchVariantVersioned::class => Versioned::DRAFT]); to the custom code MySolrIndex::init() resolved

Is this the clear approach to take for now? Ideally that unexpected behaviour would be resolved as part of this open PR so we can get this in the upcoming release in a good state.

chillu:
> So the unexpected behaviour is due to a core issue?

No, because the module still has the concept of "variants" (versioned, subsites), and I think the that variant state hasn't been excluded. So it's very much part of resolving the issue here. Adding custom code to the project Steve is testing on is no solution, it's a workaround that won't apply for everyone else. A potential "fix" could be to document this step as a customisation that everyone needs to apply to their own indexes (through changelogs), but I'd prefer for us to find a way that does this without requiring changes in every project.

PR #292, opened 2021-01-22
MNT Update require extra to use stable versions

emteknetnz:


dhensby:
this change is causing a test failure by the looks of things

PR #293, opened 2021-02-24
FIX: Index fields for subclasses

GuySartorelli:
The superclass was being checked for the field in each cycle of the loop, prohibiting the indexing of data fields from subclasses.


dhensby:
This is causing tests to fail. If the changes to the tests are expected, can you fix them and explain why? Otherwise this will probably need to be reworked

GuySartorelli:
I don't have time to come back to this, and it was fixing a fairly specific situation that I was able to handle with project-specific code. I'm going to close it.

PR #294, opened 2021-02-24
FIX: Always store the default field to support highlighting

### LINKED ISSUE #249: Highlighting with Excerpt working only in local/dev environment
Hi Team!

Highlighting feature using the `Excerpt` property described in https://github.com/silverstripe/silverstripe-fulltextsearch/blob/3.4.2/docs/en/05_advanced_configuration.md#highlighting is only working the local development environment. Using addStoredField() method in other environments have no effect.

For example addStoredField('Content') will correctly generate the entry below in the schema which set property `stored='true'`.

```<field name='SilverStripe\CMS\Model\SiteTree_Content' type='htmltext' indexed='true' stored='true' multiValued=''/>```

The `highlight` property in the Solr response will not return data for that field in a non-local environment.

```
"highlighting": {
    "2-SilverStripe\\CMS\\Model\\SiteTree-{\"SilverStripe\\\\FullTextSearch\\\\Search\\\\Variants\\\\SearchVariantSubsites\":\"0\",\"SilverStripe\\\\FullTextSearch\\\\Search\\\\Variants\\\\SearchVariantVersioned\":\"Live\"}": {}
  }
```

In local environment, I get content 
```
"highlighting": {
    "2-SilverStripe\\CMS\\Model\\SiteTree-{\"SilverStripe\\\\FullTextSearch\\\\Search\\\\Variants\\\\SearchVariantSubsites\":\"0\",\"SilverStripe\\\\FullTextSearch\\\\Search\\\\Variants\\\\SearchVariantVersioned\":\"Live\"}": {
      "_text": [
        "<p>You can fill this page out with your own content, or delete it and create your own pages. <em>some search term</em></p>"
      ]
    }
  },
```

It looks like the cause of this is the `stored` property for `_text` field in the schema is not set to 'true' except in local by the getStoredDefault() method ( https://github.com/silverstripe/silverstripe-fulltextsearch/blob/3.4.2/src/Solr/SolrIndex.php#L388-L391), which sets `stored="true"` for all fields. And this is the only way `stored` can be set for that field https://github.com/silverstripe/silverstripe-fulltextsearch/blob/3.4.2/src/Solr/SolrIndex.php#L202.

As a test in non local environment, I manually set `store="true""` for `_text` in the schema, reloaded the core, and re-indexed, I get data in `Excerpt` property, the same as the local environment.

It look like individual field values can't be configured for highlighting if the text field values are collected in `_text`?

It has affected a client project. 

Tested on:
 CMS/Framework `4.3.3` base install and fulltextsearch `3.4.2`,
(Which is why a couple of the links is referencing that tag, but looking at master, the SolrIndex code is pretty much the same.)

Cheers.

GuySartorelli:
Fixes #249

### DIFF
--- a/src/Solr/SolrIndex.php
+++ b/src/Solr/SolrIndex.php
@@ -203,7 +203,7 @@ abstract class SolrIndex extends SearchIndex
         // Add the fulltext collation field
 
         $df = $this->getDefaultField();
-        $xml[] = "<field name='{$df}' type='htmltext' indexed='true' stored='{$stored}' multiValued='true' />" ;
+        $xml[] = "<field name='{$df}' type='htmltext' indexed='true' stored='true' multiValued='true' />" ;
 
         // Add the user-specified fields
 


dhensby (changes requested):
This isn't the right fix. `getStoredDefault()` is where the value is drawn from and that has the environment type check.

I don't know why it's true for dev envs only, but that's where the logic needs to be fixed

michalkleiner:
@dhensby `getStoredDefault` provides the information to all field type definitions, not just the default one. The logic there can stay as is, but for the default field, to enable highlighting, it needs to be stored regardless of the environment type, so changing it in this single place makes sense, unless there's another method introduces just for storing the default field.

My guess why it's stored for dev env only is to inspect the indexed records' content which you may not need in test/live environments.

dhensby (approved):

GuySartorelli:
@michalkleiner That's exactly what I was thinking. Thanks for putting it into words :)
There is an unfortunately lack of lexical distinction here between the default search field and the default value for 'stored'... could that be worth raising to avoid this sort of confusion going forward?

PR #295, opened 2021-04-14
[CMS3] Add POST config store

andrewandante:
Backporting this fix to SS3 for those who might need it - happy to close if we aren't accepting submissions for SS3 anymore


NightJar:
I'm OK with this, just please do not expect a release :)
[How it is](https://github.com/silverstripe/silverstripe-fulltextsearch/compare/2.5.0...2), is how it is.

Issue #260, opened 2019-07-18
Post and WebDav Index Stores are requiring 'remotepath' and `path` when it should be optional


brettt89:
https://github.com/silverstripe/silverstripe-fulltextsearch/blob/3/src/Solr/Stores/SolrConfigStore_Post.php#L33

Both `remotepath` and `path` are mandatory in Post and WebDav Stores, however this flow does not make sense.

`path` is a variable used to define a default "base path" for the Solr server (e.g. /solr/)
 - This is used for when Solr servers are hosted on servers in which multiple environments are used on a single domain and are split out using URI.
`remotepath` seems to be an additional URI field which is also used in the same way.

Either one of these should be removed, or make one mandatory and the other optional.

e.g.

```
        if (isset($config['remotepath'])) {
            $this->remote = $config['remotepath'];
        }
```

Cheddam:
More discussion on a long term change in behaviour might be warranted here, but a basic fix that makes `remotepath` optional in `SolrConfigStore_Post` has been applied via #261.

michalkleiner:
`remotepath`, `path`, `auth` and `port` are now optional since 3.8.1

I'll close this now but feel free to raise a new issue if required.

PR #297, opened 2021-08-23
FIX Only use values that are not empty

michalkleiner:
Previously the `isset` check would still allow falsey values.

Fix for #260. It's quite possible the module won't get too much longterm love going forward but this could be seen as maintenance to allow for configurations where these settings are set to null but still incorrectly applied.


emteknetnz (approved):
Looks good.  `!empty($a[$k])` is basically shorthand for `isset($a[$k]) && $a[$k]`

emteknetnz:
I've tagged 3.8.1 https://github.com/silverstripe/silverstripe-fulltextsearch/releases/tag/3.8.1

PR #298, opened 2021-09-13
Make SearchQuery_Range injectable.

GuySartorelli:
[The docs](https://github.com/silverstripe/silverstripe-fulltextsearch/blob/3/docs/en/04_querying.md#searching-value-ranges) indicate that this class should be injectable.


michalkleiner:
Makes sense, yep. Thanks!

PR #301, opened 2021-12-01
DEP Allow symfony 4

emteknetnz:
Issue https://github.com/silverstripeltd/product-issues/issues/492

### DIFF
--- a/src/Solr/Reindex/Handlers/SolrReindexImmediateHandler.php
+++ b/src/Solr/Reindex/Handlers/SolrReindexImmediateHandler.php
@@ -95,9 +95,16 @@ class SolrReindexImmediateHandler extends SolrReindexBase
         $scriptPath = sprintf("%s%scli-script.php", $frameworkPath, DIRECTORY_SEPARATOR);
         $scriptTask = "{$php} {$scriptPath} dev/tasks/{$taskName}";
 
-        $cmd = "{$scriptTask} index={$indexClassEscaped} class={$class} group={$group} groups={$groups} variantstate={$statevar}";
-        $cmd .= " verbose=1";
-        $logger->info("Running '$cmd'");
+        $cmd = [
+            $scriptTask,
+            "index={$indexClassEscaped}",
+            "class={$class}",
+            "group={$group}",
+            "groups={$groups}",
+            "variantstate={$statevar}",
+            "verbose=1"
+        ];
+        $logger->info('Running ' . implode(' ', $cmd));
 
         // Execute script via shell
         $process = new Process($cmd);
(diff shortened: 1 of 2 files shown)


maxime-rainville (changes requested):
The [`Process`](https://github.com/silverstripe/silverstripe-fulltextsearch/blob/9de2764faaa63cb3acec9c8a5f249303cb756cec/src/Solr/Reindex/Handlers/SolrReindexImmediateHandler.php#L103) constructor now only accepts an array. Passing the command as a string will [throw a deprecation warning](https://github.com/symfony/process/blob/c2098705326addae6e6742151dfade47ac71da1b/Process.php#L146-L148).

emteknetnz:
Updated it to pass an array

**symfony 3 array|string**
https://github.com/symfony/process/blob/3.4/Process.php#L135

**symfony 4 array**
https://github.com/symfony/process/blob/4.4/Process.php

**How symfony 3 handles arrays (implodes them)**
https://github.com/symfony/process/blob/3.4/Process.php#L285

src/Solr/Reindex/Handlers/SolrReindexImmediateHandler.php
```diff
@@ -100,7 +100,7 @@ protected function processGroup(
         $logger->info("Running '$cmd'");
 
         // Execute script via shell
-        $process = new Process($cmd);
+        $process = new Process([$cmd]);
```
maxime-rainville, line 103: The [Symphony doc](https://symfony.com/doc/current/components/process.html#usage) says to pass the parameters as array entries:
```php
$process = new Process(['ls', '-lsa']);
```

So it probably needs to be something like this:
```suggestion
        $process = new Process([
            $scriptTask, 
            "index={$indexClassEscaped}",
            "class={$class}",
            "group={$group}",
            "groups={$groups}",
            "variantstate={$statevar}",
            "verbose=1"
        ]);
```

maxime-rainville (changes requested):

src/Solr/Reindex/Handlers/SolrReindexImmediateHandler.php
emteknetnz: Yes agree that would be better, have updated

maxime-rainville (approved):
Tested locally with both symfony 3 and 4. Works great.

GuySartorelli:
@maxime-rainville Heya, just wondering if this got tested in dev mode, as well as test/live mode? I'm having an issue at the moment after an update that includes updating symfony from 3 to 4 - in dev mode the processes don't seem to be run, though replacing that with `shell_exec` does work.

Issue #302, opened 2021-12-13
Migrate to new Symfony cache classes


emteknetnz:
```Exception: HTML Parse Error: The "Symfony\Component\Cache\Simple\PhpFilesCache" class is deprecated since Symfony 4.3, use "Symfony\Component\Cache\Adapter\PhpFilesAdapter" and type-hint for "Symfony\Contracts\Cache\CacheInterface" instead. in /sites/cmstest-uat/releases/20211213214917/vendor/silverstripe/fulltextsearch/src/Search/Indexes/SearchIndex.php:502 Stack trace: #0 [internal function]: SilverStripe\FullTextSearch\Search\Indexes\SearchIndex->SilverStripe\FullTextSearch\Search\Indexes\{closure}(16384, 'The "Symfony\\Co...', '/sites/cmstest-...', 20, Array)```

Saw warning on queuedjob message when manually running solr reindex job

I think it's blocking the indexing of content block content on AWS, possibly only when running PHP 8.0

## Acceptance criteria
- [ ] Warnings are visible in the output with failing the indexing tasks
- [ ] The deprecated Symfony API is not used anymore
- [ ] No module reference the old cache implementation
- [ ] All modules that require symphony/cache, are constrained to using the relevant version of Symphony and the kitchen sink is installable.
- [ ] The Fulltext Search cache still works.

### PRs
- [x] https://github.com/silverstripe/silverstripe-fulltextsearch/pull/303
- [ ] https://github.com/silverstripe/silverstripe-framework/pull/10187
- [ ] https://github.com/silverstripe/silverstripe-versioned/pull/349

**Comment by emteknetnz** [comment]
Error comes from

*PhpFilesCache.php*

```
namespace Symfony\Component\Cache\Simple;
...
`@trigger_error(sprintf('The "%s" class is deprecated since Symfony 4.3, use "%s" and type-hint for "%s" instead.', PhpFilesCache::class, PhpFilesAdapter::class, CacheInterface::class), \E_USER_DEPRECATED);`
...
class PhpFilesCache extends AbstractCache implements PruneableInterface
```[](url)

michalkleiner:
Are we ok just with silencing this when getting the field content?

emteknetnz:
Yes. I've got a PR up, will test on an environment that was throwing the error shortly

maxime-rainville:
Merged the PR as a stop gap. But it doesn't address the fundamental issue.

[...]

PR #303, opened 2021-12-14
FIX Allow deprecated warnings

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-fulltextsearch/issues/302

### DIFF
--- a/src/Search/Indexes/SearchIndex.php
+++ b/src/Search/Indexes/SearchIndex.php
@@ -501,7 +501,7 @@ abstract class SearchIndex extends ViewableData
         $errorHandler = function ($no, $str) {
             throw new Exception('HTML Parse Error: ' . $str);
         };
-        set_error_handler($errorHandler, E_ALL);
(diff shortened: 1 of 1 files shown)


michalkleiner (approved):

src/Search/Indexes/SearchIndex.php
```diff
@@ -501,7 +501,7 @@ protected function _getFieldValue($object, $field)
         $errorHandler = function ($no, $str) {
             throw new Exception('HTML Parse Error: ' . $str);
         };
-        set_error_handler($errorHandler, E_ALL);
+        set_error_handler($errorHandler, E_ALL & ~(E_DEPRECATED | E_USER_DEPRECATED));
```
maxime-rainville, line 504: That `$errorHandler` is weak ... we need to make it smarter so we don't just suppress deprecation warning.

maxime-rainville (approved):
That solution is OK as a stop-gap measure. But it doesn't adress tho fundamental problem that the error handling logic is dumb.

This doesn't fix the issue, but it unblocks the release for now.

PR #306, opened 2022-01-18
FIX Don't assume DataObject::canView always returns bool

### LINKED ISSUE #304: `SearchableService::isSearchable` incorrectly assumes return type of `canView`
[This method](https://github.com/silverstripe/silverstripe-fulltextsearch/blob/3/src/Search/Services/SearchableService.php#L127) type-hints a boolean, but also returns the value of `DataObject::canView` which is a tri-state. 

Either the tri-state should be dealt with, or the value should be casted.

GuySartorelli:
Fixes #304 
See also silverstripe/silverstripe-elemental#951

Because there is no return value typehinting in DataObject::canView, the value returned from that method can be of any type. We must cast to boolean before returning the value to avoid possible errors with non-boolean return types.

### DIFF
--- a/src/Search/Services/SearchableService.php
+++ b/src/Search/Services/SearchableService.php
@@ -151,12 +151,12 @@ class SearchableService
                 // Anonymous member canView() for indexing
                 if (!$this->classSkipsCanViewCheck($objClass)) {
                     $value = Member::actAs(null, function () use ($obj) {
-                        return $obj->canView();
+                        return (bool) $obj->canView();
                     });
                 }
             } else {
                 // Current member canView() check for retrieving search results
-                $value = $obj->canView();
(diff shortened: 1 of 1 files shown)


michalkleiner (approved):
I'd probably target 3.8 since it's technically a bug fix and 3.9 is not out yet.

GuySartorelli:
@michalkleiner I avoid targetting specific versions because I have seen far too many PRs targetted for a specific version with 2+ comments afterward asking to retarget because it never got merged in.

emteknetnz:
@GuySartorelli I've re-targeted to 3.9 and released a patch version for you - [3.9.1](https://github.com/silverstripe/silverstripe-fulltextsearch/releases/tag/3.9.1)

PR #315, opened 2022-03-02
[Backport] Let symfony/process escape each command part separately

michalkleiner:
Previously the executable, the PHP script and the task were a single string which `symfony/process` tried to wrap in apostrophes which subsequently failed as an invalid executable with error code 127 "command not found".

When each part of the command including the php, the CLI script and the task name, is passed in separately, the process does escape it correctly.

### DIFF
--- a/src/Solr/Reindex/Handlers/SolrReindexImmediateHandler.php
+++ b/src/Solr/Reindex/Handlers/SolrReindexImmediateHandler.php
@@ -75,17 +75,12 @@ class SolrReindexImmediateHandler extends SolrReindexBase
         $taskName
     ) {
         $indexClass = get_class($indexInstance);
-
-        // Build script parameters
-        $indexClassEscaped = $indexClass;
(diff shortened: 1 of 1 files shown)


GuySartorelli:
@michalkleiner Is this just a port of #309 to 3.9.x?

michalkleiner:
Hah, well, it may seem that way, though I arrived at it through my own painful troubleshooting.

michalkleiner:
#309 probably should've targetted the minor as it was a fix already back then. Would also save me time.

michalkleiner:
@emteknetnz could you please give this a tick so that we can release 3.9.4?

GuySartorelli:
> #309 probably should've targetted the minor as it was a fix already back then. Would also save me time.
You're not wrong.

emteknetnz:
Done https://github.com/silverstripe/silverstripe-fulltextsearch/releases/tag/3.9.4

PR #328, opened 2022-07-21
MNT Fix unit test

emteknetnz:
Issue https://github.com/silverstripe/gha-ci/issues/37

Might fix https://github.com/silverstripe/recipe-solr-search/runs/7459144288?check_suite_focus=true#step:12:69
`1) SilverStripe\FullTextSearch\Tests\BatchedProcessorTest::testBatching
Failed asserting that 15 matches expected 9.`

Don't need to release, though ensure is merged all the way up as recipe-solr-search uses x-dev branches


GuySartorelli (approved):
Regardless of whether this actually fixes the test, it is more correct and therefore a good change to make.

PR #349, opened 2023-05-10
FIX Reset class from the actual SearchVariant instance to respect Injector

### LINKED ISSUE #348: `SearchVariant` replaced by `Injector` not consistent when used as part of `_documentid`
Still trying to figure out what is going on, but the gist is that we are replacing a sublass of SearchVariant with a custom subclass of it via Injector to add an extension point to it.

When checking Solr log, documents added to or updated in the solr index have `_documentid` field containing both the original and the replace variant as part of the id whereas the delete action does not have the original variant in the id, effectively causing the documents not being removed from Solr index on an unpublish action. The front-end still does `canView` check before displaying the results, so this doesn't cause a problem in terms of visibility of stale search results, but in places where a set number of results is expected it may mean Solr e.g. returns 3 but FE only displays 2 items.

## PRs
- https://github.com/silverstripe/silverstripe-fulltextsearch/pull/349

michalkleiner:
Fixes #348 although SearchVariant doesn't explicitly use `Injectable` trait, it uses `ClassInfo` and then `Injector` to create instances of each variant, which can be impacted by `Injector` and produce incorrect list of variants to be applied.

This is a no-harm tweak that addresses this (probably rare) use case.

### DIFF
--- a/src/Search/Variants/SearchVariant.php
+++ b/src/Search/Variants/SearchVariant.php
@@ -105,6 +105,9 @@ abstract class SearchVariant
                     $ref = new ReflectionClass($variantclass);
                     if ($ref->isInstantiable()) {
                         $variant = singleton($variantclass);
+                        // reassign actual class since Injector could be involved when creating the singleton
+                        $variantclass = get_class($variant);
+
                         if ($variant->appliesToEnvironment()) {
                             $concrete[$variantclass] = $variant;
                         }


GuySartorelli (commented):
Does `ClassInfo::subclassesFor()` give you both the injected class _and_ the class being replaced in its list? If so, do we need to dedupe that?

michalkleiner:
Yes, it gives both classes since it just gets it from native PHP inheritance, it doesn't work with the Injector replacement concept at that point in time. So the key is the PHP class and can't be deduped easily. It gets deduped as a byproduct when the list of instances is created (with my change) where the key is changed to the implementing class respecting Injector.

I can see if I can do it in some other way, dedupe it before the instances are created, or possibly update `ClassInfo::subclassesFor` to respect `Injector` replacements, but that would go to 5.x and unless backported to 4.13 would still be unusable for us on CMS 4.

GuySartorelli:
I was thinking more along the lines of double checking if we've already processed that class in the loop, e.g:

```php
// Don't use the original class if it was already added by some injected replacement
if ($variantClass === get_class($variant) && isset($concrete[$variantclass])) {
    continue; // because we already handled the replacement
}
```

It's probably an even edgier edge case than what you initially encountered, as I think for that condition to be true the injected replacement has to _not_ be a subclass of the original variant it's replacing...

michalkleiner:
Right, yeah, the mechanism works on the premise that it specifically looks for subclasses of `SearchVariant`, so yeah, replacing it with with something that is not a subclass of it could have even more issues with it (if it was working at all).

michalkleiner:
But happy to adjust as needed as long as it's deduped in the end which is the issue we needed resolved.

GuySartorelli:
The replacement could be a subclass of `SearchVariant` but not a subclass of the _specific subclass of `SearchVariant` that's being replaced_, is what I'm getting at. I'm not overly fussed about that edge case though - that's super unlikely.

I'll let you respond, but if I see no changes or response by the end of the day I think I'll merge as-is as this solves the specific problem it was intended to solve and doesn't cause any obvious regressions.

michalkleiner:
Gotcha. I think that's quite a stretch for now so won't be doing anything more in that direction. So if there's nothing else we can merge this — appreciate the quick review!

GuySartorelli (approved):

GuySartorelli:
Tagged as [3.12.1](https://github.com/silverstripe/silverstripe-fulltextsearch/releases/tag/3.12.1)

PR #351, opened 2023-08-09
FIX Remove call to process::inheritEnvironmentVariables()

LeviWB:
inheritEnvironmentVariables() was removed from the Symfony Process component in 5.0.0, as enviroment variables are always inherited. (See https://github.com/symfony/process/blob/5.0/CHANGELOG.md)

Previously I was getting a `[Emergency] Uncaught Error: Call to undefined method Symfony\Component\Process\Process::inheritEnvironmentVariables()` when the immediate reindex handler was called (e.g. when a reindex was run while the site is in dev mode).


GuySartorelli (approved):
Makes sense. Thanks for submitting this. Will merge on green

GuySartorelli:
This will be automatically tagged once CI finishes running on the branch.

PR #359, opened 2025-12-04
Upgrade/cms6 compatibility

satrun77:
## Issues

- #

## Pull request checklist

- [ ] The target branch is correct
    - See [picking the right version](https://docs.silverstripe.org/en/contributing/code/#picking-the-right-version)
- [ ] All commits are relevant to the purpose of the PR (e.g. no debug statements, unrelated refactoring, or arbitrary linting)
    - Small amounts of additional linting are usually okay, but if it makes it hard to concentrate on the relevant changes, ask for the unrelated changes to be reverted, and submitted as a separate PR.
- [ ] The commit messages follow our [commit message guidelines](https://docs.silverstripe.org/en/contributing/code/#commit-messages)
- [ ] The PR follows our [contribution guidelines](https://docs.silverstripe.org/en/contributing/code/)
- [ ] Code changes follow our [coding conventions](https://docs.silverstripe.org/en/contributing/coding_conventions/)
- [ ] This change is covered with tests (or tests aren't necessary for this change)
- [ ] Any relevant User Help/Developer documentation is updated; for impactful changes, information is added to the changelog for the intended release
- [ ] CI is green


src/Search/Indexes/SearchIndex.php
satrun77, line 1: @sgray-circle8 this class make use of a method that no longer exists `castingClass` 

https://docs.silverstripe.org/en/5/changelogs/5.4.0/

The code in CMS5

```
    public function castingClass($field)
    {
        // Strip arguments
        $spec = $this->castingHelper($field); // <-- this still exists we can call this and then in the current class do the following line
        return trim(strtok($spec ?? '', '(') ?? '');
    }
```

satrun77:
@sgray-circle8 deprecation errors from dev/build

```
PHP Deprecated:  SilverStripe\FullTextSearch\Search\Criteria\SearchCriteria::__construct(): Implicitly marking parameter $searchQueryWriter as nullable is deprecated, the explicit nullable type must be used instead in vendor/silverstripe/fulltextsearch/src/Search/Criteria/SearchCriteria.php on line 59

Deprecated: SilverStripe\FullTextSearch\Search\Criteria\SearchCriteria::__construct(): Implicitly marking parameter $searchQueryWriter as nullable is deprecated, the explicit nullable type must be used instead in vendor/silverstripe/fulltextsearch/src/Search/Criteria/SearchCriteria.php on line 59
PHP Deprecated:  SilverStripe\FullTextSearch\Search\Criteria\SearchCriteria::create(): Implicitly marking parameter $searchQueryWriter as nullable is deprecated, the explicit nullable type must be used instead in vendor/silverstripe/fulltextsearch/src/Search/Criteria/SearchCriteria.php on line 77

Deprecated: SilverStripe\FullTextSearch\Search\Criteria\SearchCriteria::create(): Implicitly marking parameter $searchQueryWriter as nullable is deprecated, the explicit nullable type must be used instead in vendor/silverstripe/fulltextsearch/src/Search/Criteria/SearchCriteria.php on line 77
PHP Deprecated:  SilverStripe\FullTextSearch\Search\Criteria\SearchCriteria::addAnd(): Implicitly marking parameter $searchQueryWriter as nullable is deprecated, the explicit nullable type must be used instead in vendor/silverstripe/fulltextsearch/src/Search/Criteria/SearchCriteria.php on line 140

Deprecated: SilverStripe\FullTextSearch\Search\Criteria\SearchCriteria::addAnd(): Implicitly marking parameter $searchQueryWriter as nullable is deprecated, the explicit nullable type must be used instead in vendor/silverstripe/fulltextsearch/src/Search/Criteria/SearchCriteria.php on line 140
PHP Deprecated:  SilverStripe\FullTextSearch\Search\Criteria\SearchCriteria::addOr(): Implicitly marking parameter $searchQueryWriter as nullable is deprecated, the explicit nullable type must be used instead in vendor/silverstripe/fulltextsearch/src/Search/Criteria/SearchCriteria.php on line 161

Deprecated: SilverStripe\FullTextSearch\Search\Criteria\SearchCriteria::addOr(): Implicitly marking parameter $searchQueryWriter as nullable is deprecated, the explicit nullable type must be used instead in vendor/silverstripe/fulltextsearch/src/Search/Criteria/SearchCriteria.php on line 161
PHP Deprecated:  SilverStripe\FullTextSearch\Search\Criteria\SearchCriteria::getCriterionForCondition(): Implicitly marking parameter $searchQueryWriter as nullable is deprecated, the explicit nullable type must be used instead in vendor/silverstripe/fulltextsearch/src/Search/Criteria/SearchCriteria.php on line 182

Deprecated: SilverStripe\FullTextSearch\Search\Criteria\SearchCriteria::getCriterionForCondition(): Implicitly marking parameter $searchQueryWriter as nullable is deprecated, the explicit nullable type must be used instead in vendor/silverstripe/fulltextsearch/src/Search/Criteria/SearchCriteria.php on line 182
PHP Deprecated:  SilverStripe\FullTextSearch\Search\Criteria\SearchCriterion::__construct(): Implicitly marking parameter $searchQueryWriter as nullable is deprecated, the explicit nullable type must be used instead in vendor/silverstripe/fulltextsearch/src/Search/Criteria/SearchCriterion.php on line 128

Deprecated: SilverStripe\FullTextSearch\Search\Criteria\SearchCriterion::__construct(): Implicitly marking parameter $searchQueryWriter as nullable is deprecated, the explicit nullable type must be used instead in vendor/silverstripe/fulltextsearch/src/Search/Criteria/SearchCriterion.php on line 128
PHP Deprecated:  SilverStripe\FullTextSearch\Search\Queries\SearchQuery::filterBy(): Implicitly marking parameter $searchQueryWriter as nullable is deprecated, the explicit nullable type must be used instead in vendor/silverstripe/fulltextsearch/src/Search/Queries/SearchQuery.php on line 208

Deprecated: SilverStripe\FullTextSearch\Search\Queries\SearchQuery::filterBy(): Implicitly marking parameter $searchQueryWriter as nullable is deprecated, the explicit nullable type must be used instead in vendor/silverstripe/fulltextsearch/src/Search/Queries/SearchQuery.php on line 208
PHP Deprecated:  SilverStripe\FullTextSearch\Solr\Forms\SearchForm::__construct(): Implicitly marking parameter $controller as nullable is deprecated, the explicit nullable type must be used instead in vendor/silverstripe/fulltextsearch/src/Solr/Forms/SearchForm.php on line 30

Deprecated: SilverStripe\FullTextSearch\Solr\Forms\SearchForm::__construct(): Implicitly marking parameter $controller as nullable is deprecated, the explicit nullable type must be used instead in vendor/silverstripe/fulltextsearch/src/Solr/Forms/SearchForm.php on line 30
PHP Deprecated:  SilverStripe\FullTextSearch\Solr\Forms\SearchForm::__construct(): Implicitly marking parameter $fields as nullable is deprecated, the explicit nullable type must be used instead in vendor/silverstripe/fulltextsearch/src/Solr/Forms/SearchForm.php on line 30

Deprecated: SilverStripe\FullTextSearch\Solr\Forms\SearchForm::__construct(): Implicitly marking parameter $fields as nullable is deprecated, the explicit nullable type must be used instead in vendor/silverstripe/fulltextsearch/src/Solr/Forms/SearchForm.php on line 30
PHP Deprecated:  SilverStripe\FullTextSearch\Solr\Forms\SearchForm::__construct(): Implicitly marking parameter $actions as nullable is deprecated, the explicit nullable type must be used instead in vendor/silverstripe/fulltextsearch/src/Solr/Forms/SearchForm.php on line 30

Deprecated: SilverStripe\FullTextSearch\Solr\Forms\SearchForm::__construct(): Implicitly marking parameter $actions as nullable is deprecated, the explicit nullable type must be used instead in vendor/silverstripe/fulltextsearch/src/Solr/Forms/SearchForm.php on line 30
```

src/Solr/Stores/SolrConfigStore_File.php
atr887: PHP 8.2 deprecates dynamic properties. Adding explicit property declarations here avoids deprecation warning

```suggestion

        public string $local;
        public string $remote;
        
```

atr887 (changes requested):

src/Solr/Stores/SolrConfigStore_File.php
```diff
@@ -11,7 +11,10 @@
  * by just using file copies
  */
 class SolrConfigStore_File implements SolrConfigStore
-{
```
atr887, line 12: Opening brace accidentally removed

```suggestion
{
```

atr887 (changes requested):

src/Solr/Stores/SolrConfigStore_File.php
satrun77: Fixed

atr887:
Hiya, just wondering if there is an estimate of when this might be merged into master for V6 compatibility? 

thanks

satrun77:
> Hiya, just wondering if there is an estimate of when this might be merged into master for V6 compatibility?
> 
> thanks
This module is no longer maintained. The PR contains very rough code for anyone who may need it. I’ll ask about merging the PR.

src/Solr/Tasks/Solr_Reindex.php
```diff
@@ -97,21 +106,32 @@ protected function doReindex($request)
...
+    }
+
+    public function getOptions(): array
+    {
+        return [
+            new InputOption('class', null, InputOption::VALUE_NONE, 'Re-index specific class'),
+            new InputOption('index', null, InputOption::VALUE_NONE, 'Reindex specific index'),
+            new InputOption('groups', null, InputOption::VALUE_NONE, 'Groups ID'),
+            new InputOption('group', null, InputOption::VALUE_NONE, 'Group ID'),
+            new InputOption('variantstate', null, InputOption::VALUE_NONE, 'variantstate'),
```
blueo: These are all declared `VALUE_NONE` (boolean flags that take no value), but `doReindex()` reads them as values via `$input->getOption('index')`, `getOption('class')` etc.

Reproduced on this branch — the task is unusable from the CLI:

```
$ vendor/bin/sake tasks:solr-reindex --index=SolrIndexTest_FakeIndex

  The "--index" option does not accept a value.

$ vendor/bin/sake tasks:solr-reindex --class=Page

  The "--class" option does not accept a value.
```

`--help` shows the same thing — the options render as bare flags rather than taking a value:

```
Options:
      --class           Re-index specific class
      --index           Reindex specific index
      --groups          Groups ID
      --group           Group ID
      --variantstate    variantstate
```

Worth noting why CI doesn't catch this: the only caller exercised by the test suite is `SolrReindexImmediateHandler::executeBuiltTask()`, which builds an `ArrayInput`. `ArrayInput` doesn't validate values against the option definition, so it passes them straight through and the tests stay green. Only the `ArgvInput` path that `sake` uses enforces it.

These should be `InputOption::VALUE_REQUIRED` (or `VALUE_OPTIONAL`), and it'd be worth a smoke test that actually goes through the CLI definition.

src/Solr/Tasks/Solr_BuildTask.php
```diff
@@ -1,69 +1,43 @@
...
-    /**
-     * Assign a new logger
-     *
-     * @param LoggerInterface $logger
-     */
-    public function setLogger(LoggerInterface $logger)
-    {
-        $this->logger = $logger;
-    }
+    protected bool $is_enabled = false;
```
blueo: `is_enabled` needs to be `private static` to be picked up by the Config layer — that's how `BuildTask::isEnabled()` reads it:

```php
// framework/src/Dev/BuildTask.php
private static bool $is_enabled = true;

public function isEnabled(): bool
{
    return $this->config()->get('is_enabled');
}
```

As a `protected` instance property this is invisible to Config, so it has no effect. Confirmed by reflection on the built classes:

```
Solr_Configure: is_enabled static=false visibility=protected
Solr_Reindex:   is_enabled static=false visibility=protected
```

Both tasks do end up enabled, but only because they inherit the framework default of `true` — not because of these declarations. Suggest `private static bool $is_enabled = ...;` in all three classes.

src/Solr/Tasks/Solr_Configure.php
```diff
@@ -9,17 +11,27 @@
...
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
 
 class Solr_Configure extends Solr_BuildTask
 {
-    private static $segment = 'Solr_Configure';
-    protected $enabled = true;
 
-    public function run($request)
+    protected bool $is_enabled = true;
```
blueo: Same as the `is_enabled` note in `Solr_BuildTask` — needs to be `private static` to be read as config.

src/Solr/Tasks/Solr_Reindex.php
```diff
@@ -27,17 +30,21 @@
  */
 class Solr_Reindex extends Solr_BuildTask
 {
-    private static $segment = 'Solr_Reindex';
+    protected bool $is_enabled = true;
```
blueo: Same as the `is_enabled` note in `Solr_BuildTask` — needs to be `private static` to be read as config.

src/Solr/Tasks/Solr_Configure.php
```diff
@@ -30,17 +42,17 @@ public function run($request)
                 $this->updateIndex($instance, $store);
             } catch (Exception $e) {
                 // We got an exception. Warn, but continue to next index.
-                $this
-                    ->getLogger()
-                    ->error("Failure: " . $e->getMessage());
+                $this->info("Failure: " . $e->getMessage());
```
blueo: This potentially hiding more logs than intentional. `Solr_BuildTask::info()` defaults to `$hidden = true`, so unless `--verbose` is passed nothing is written. Previous code called `$this->getLogger()->error(...)`, which would always print. 
fine to ignore if intended behaviour

src/Solr/Reindex/Handlers/SolrReindexImmediateHandler.php
```diff
@@ -86,29 +94,32 @@ protected function processGroup(
...
-        // Execute script via shell
-        $process = new Process($cmd);
-
-        // Set timeout from config. Process default is 60 seconds
-        $process->setTimeout($this->config()->get('process_timeout'));
-
-        $process->run();
-
-        $res = $process->getOutput();
+        $logger->writeln('Running ' . implode(' ', $cmd));
```
blueo: `$cmd` is built and logged here but never executed the work now happens in `executeBuiltTask()` below, which runs the task in the same process.

That's a behavioural change as the original spawned a separate PHP process per group so that a long reindex couldn't accumulate memory across groups; the class docblock still says as much:

> Internally batches of records will be invoked via shell tasks in the background

this means:

- long reindexes now run entirely in one process (memory growth across all groups)
- `process_timeout` is no longer applied to anything
- the `php_bin` config, `$php`, `$frameworkPath`, `$scriptPath` and the `Symfony\Component\Process\Process` import are all now dead
- the docblock at the top of the class is inaccurate

If running in-process is intentional (CMS 6 dropped `cli-script.php`, so the old invocation had to change), that's fine but the dead `$cmd`, the stale config and the docblock should be cleaned up, and it's worth a note in the PR description since it changes the memory profile of a full reindex.

src/Solr/Reindex/Handlers/SolrReindexImmediateHandler.php
```diff
@@ -86,29 +94,32 @@ protected function processGroup(
...
+                '--index' => $indexClass,
+                '--class' => $class,
+                '--group' => $group,
+                '--groups' => $groups,
+                '--variantstate' => $statevar,
+                '--verbose' => true,
+            ],
+            true
+        );
         if ($logger) {
```
blueo: `$logger` is a non-nullable `PolyOutput` parameter so it should always be true

src/Solr/Reindex/Handlers/SolrReindexImmediateHandler.php
```diff
@@ -119,4 +130,27 @@ protected function processGroup(
         // This will slow down things a tiny bit, but it is done so that we don't timeout to the database during a reindex
         DB::query('SELECT 1');
     }
+
+    public function executeBuiltTask(string $className, array $params = [], bool $returnOutput = false): ?string
+    {
+        $definition = [];
+        $paramNames = array_keys($params);
```
blueo: `$definition` and `$paramNames` are assigned but never read

tests/SolrReindexTest.php
```diff
@@ -247,15 +250,15 @@ public function testRunGroup()
...
         $this->getHandler()->runGroup($logger, $this->index, $state, SolrReindexTest_Item::class, 6, 2);
         $idMessage = $logger->filterMessages('Updated ');
         $this->assertNotEmpty(preg_match('/^Updated (?<ids>[,\d]+)/i', $idMessage[0] ?? '', $matches));
         $ids = array_unique(explode(',', $matches['ids'] ?? ''));
 
         // Test successful
-        $this->assertNotEmpty($logger->getMessages('Adding ' . SolrReindexTest_Item::class));
-        $this->assertNotEmpty($logger->getMessages('Done'));
+        $this->assertNotEmpty($logger->getMessages());
+        $this->assertNotEmpty($logger->getMessages());
```
blueo: These two assertions are now identical and only prove that *something* was logged — the specific behaviour they were checking ("Adding <class>" and "Done" were emitted) is no longer covered.

In fairness the originals were already broken as`getMessages()` never took an argument. But the method you want does exist and is used a few lines above at `filterMessages('Updated ')`:

```php
$this->assertNotEmpty($logger->filterMessages('Adding ' . SolrReindexTest_Item::class));
$this->assertNotEmpty($logger->filterMessages('Done'));
```

I ran that locally against this branch and it passes (`OK (2 tests, 40 assertions)`) — so the logging behaviour survived the `PolyOutput` migration intact, and this is a chance to make the assertions real rather than drop them.

tests/SolrReindexQueuedTest.php
```diff
@@ -186,12 +187,10 @@ public function testReindexSegmentsGroups()
...
 
         // Test that invalid classes are removed
-        $this->assertNotEmpty($logger->getMessages('Clearing obsolete classes from ' . SolrReindexTest_Index::class));
+        $this->assertNotEmpty($logger->getMessages());
 
         // Test that valid classes in invalid variants are removed
-        $this->assertNotEmpty($logger->getMessages(
-            'Clearing all records of type ' . SolrReindexTest_Item::class . ' in the current state: {"' . SolrReindexTest_Variant::class . '":"2"}'
-        ));
+        $this->assertNotEmpty($logger->getMessages());
```
blueo: Same as in `SolrReindexTest` — these two are identical and assert only that some message exists. `filterMessages()` will restore the original intent:

```php
$this->assertNotEmpty($logger->filterMessages('Clearing obsolete classes from ' . SolrReindexTest_Index::class));
$this->assertNotEmpty($logger->filterMessages(
    'Clearing all records of type ' . SolrReindexTest_Item::class . ' in the current state: ...'
));
```

tests/Traits/AssertsConsecutiveCalls.php
```diff
@@ -0,0 +1,32 @@
...
+     * use `expects($this->exactly(n))` if the number of calls matters.
+     *
+     * @param array $expectedArgs One entry per expected invocation
+     */
+    protected function withConsecutiveArgs(array $expectedArgs): callable
+    {
+        $remaining = $expectedArgs;
+
+        return function (mixed $actualArg) use (&$remaining): void {
+            if ($remaining === []) {
```
blueo: Worth noting this is a looser check than the `withConsecutive()` it replaces: PHPUnit would fail if the mock was called *more* times than there were expectation sets, whereas this returns early and passes silently.

The docblock calls it out and most call sites pair it with `expects($this->exactly(n))`, but `SolrReindexTest::testReindexSegmentsGroups()` doesn't — so an unexpected extra `deleteByQuery()` call would go undetected there.

src/Solr/Tasks/Solr_Reindex.php
```diff
@@ -97,21 +106,32 @@ protected function doReindex($request)
...
 
-            $handler->runGroup($this->getLogger(), $indexInstance, $state, $class, $groups, $group);
+            $handler->runGroup($this->output, $indexInstance, $state, $class, $groups, $group);
             return;
         }
 
         // If run at the top level, delegate to appropriate handler
-        $taskName = $this->config()->segment ?: get_class($this);
-        $handler->triggerReindex($this->getLogger(), $this->config()->recordsPerRequest, $taskName, $class);
+        $handler->triggerReindex($this->output, $this->config()->recordsPerRequest, self::class, $class);
```
blueo, line 126: `$taskName` used to be `$this->config()->segment ?: get_class($this)` — i.e. the string `Solr_Reindex`. It's now the FQCN. That value isn't used here, so the effect isn't visible in this hunk; it's threaded down through the handler and consumed in two different places that want two different formats.

**Where it goes (immediate handler, the default):**

1. [`Solr_Reindex::doReindex()` L124](https://github.com/silverstripe/silverstripe-fulltextsearch/blob/16985522db3fa0ad4cb001c98bbfc8a8f4e8662c/src/Solr/Tasks/Solr_Reindex.php#L124) — `$handler->triggerReindex($this->output, ..., self::class, $class)`
2. [`SolrReindexImmediateHandler::triggerReindex()` L41-44](https://github.com/silverstripe/silverstripe-fulltextsearch/blob/16985522db3fa0ad4cb001c98bbfc8a8f4e8662c/src/Solr/Reindex/Handlers/SolrReindexImmediateHandler.php#L41-L44) → `runReindex()`
3. [`SolrReindexBase::runReindex()` L22-28](https://github.com/silverstripe/silverstripe-fulltextsearch/blob/16985522db3fa0ad4cb001c98bbfc8a8f4e8662c/src/Solr/Reindex/Handlers/SolrReindexBase.php#L22-L28) → `processIndex()`
4. [`SolrReindexBase::processIndex()` L60](https://github.com/silverstripe/silverstripe-fulltextsearch/blob/16985522db3fa0ad4cb001c98bbfc8a8f4e8662c/src/Solr/Reindex/Handlers/SolrReindexBase.php#L60) → `processVariant()`
5. [`SolrReindexBase::processVariant()` L131](https://github.com/silverstripe/silverstripe-fulltextsearch/blob/16985522db3fa0ad4cb001c98bbfc8a8f4e8662c/src/Solr/Reindex/Handlers/SolrReindexBase.php#L131) → `processGroup()`
6. [`SolrReindexImmediateHandler::processGroup()` L76-83](https://github.com/silverstripe/silverstripe-fulltextsearch/blob/16985522db3fa0ad4cb001c98bbfc8a8f4e8662c/src/Solr/Reindex/Handlers/SolrReindexImmediateHandler.php#L76-L83) — two consumers:
   - [L98](https://github.com/silverstripe/silverstripe-fulltextsearch/blob/16985522db3fa0ad4cb001c98bbfc8a8f4e8662c/src/Solr/Reindex/Handlers/SolrReindexImmediateHandler.php#L98) `"tasks:{$taskName}"`, logged at [L106](https://github.com/silverstripe/silverstripe-fulltextsearch/blob/16985522db3fa0ad4cb001c98bbfc8a8f4e8662c/src/Solr/Reindex/Handlers/SolrReindexImmediateHandler.php#L106) — wants a **command name**
   - [L109-110](https://github.com/silverstripe/silverstripe-fulltextsearch/blob/16985522db3fa0ad4cb001c98bbfc8a8f4e8662c/src/Solr/Reindex/Handlers/SolrReindexImmediateHandler.php#L109-L110) `executeBuiltTask($taskName, ...)` → [L138](https://github.com/silverstripe/silverstripe-fulltextsearch/blob/16985522db3fa0ad4cb001c98bbfc8a8f4e8662c/src/Solr/Reindex/Handlers/SolrReindexImmediateHandler.php#L138) `$className::create()` — wants a **class name**

So the FQCN is correct for `executeBuiltTask()` and wrong for the log line, which now prints:

```
Running sake tasks:SilverStripe\FullTextSearch\Solr\Tasks\Solr_Reindex --index=... --class=...
```

Suggest keeping the FQCN for execution and using `self::getName()` (→ `tasks:solr-reindex`) when building the display string, so the logged command is one someone could actually paste.

**Second path worth checking — the value gets persisted.** With the queued handler:

- [`SolrReindexQueuedHandler::triggerReindex()` L79](https://github.com/silverstripe/silverstripe-fulltextsearch/blob/16985522db3fa0ad4cb001c98bbfc8a8f4e8662c/src/Solr/Reindex/Handlers/SolrReindexQueuedHandler.php#L79) → `new SolrReindexQueuedJob($batchSize, $taskName, $classes)`
- [`SolrReindexQueuedJob::getJobData()` L60](https://github.com/silverstripe/silverstripe-fulltextsearch/blob/16985522db3fa0ad4cb001c98bbfc8a8f4e8662c/src/Solr/Reindex/Jobs/SolrReindexQueuedJob.php#L60) — `$data->jobData->taskName = $this->taskName`, serialised into the `QueuedJobDescriptor` row
- restored in [`setJobData()` L73](https://github.com/silverstripe/silverstripe-fulltextsearch/blob/16985522db3fa0ad4cb001c98bbfc8a8f4e8662c/src/Solr/Reindex/Jobs/SolrReindexQueuedJob.php#L73) and replayed at [L96](https://github.com/silverstripe/silverstripe-fulltextsearch/blob/16985522db3fa0ad4cb001c98bbfc8a8f4e8662c/src/Solr/Reindex/Jobs/SolrReindexQueuedJob.php#L96)

That means any reindex job queued before the upgrade carries the old `'Solr_Reindex'` segment in its stored job data, and on replay would reach `executeBuiltTask('Solr_Reindex')` → `'Solr_Reindex'::create()` → class-not-found. Probably worth either draining the queue as an upgrade step or making `executeBuiltTask()` tolerate a non-FQCN name.

blueo (changes requested):
i had a run over with some automation and picked up a couple of things :)

satrun77:
@blueo I have addressesed your feedback

blueo (approved):
cool tested locally and seems to be working as expected ✅

satrun77:
> Hiya, just wondering if there is an estimate of when this might be merged into master for V6 compatibility?
> 
> thanks

Tagged release is now available.

atr887:
> > Hiya, just wondering if there is an estimate of when this might be merged into master for V6 compatibility?
> > thanks
> 
> Tagged release is now available.

thankyou, much appreciated.

